PR #40, opened 2026-09-23
fix: Keep the inline-image drag ghost on a ref so the React Compiler accepts it

r4iju:
Master has not deployed since #38: the Vercel build fails in `babel-plugin-react-compiler` (`compilationMode: "all"`, `panicThreshold: "critical_errors"`) on the lazy drag-image memo in `InlineImagePlugin`:

```
Modifying a variable defined outside a component or hook is not allowed. Consider using an effect.
> 348 |     dragImage.src = TRANSPARENT_IMAGE;
```

The transparent drag ghost is now created in the plugin's existing effect and held on a `useRef`, so the module still loads without a DOM (the #38 goal) and no module-scope variable is mutated.

## Verification

- `NODE_ENV=production next build --turbopack` locally: `✓ Compiled successfully`, all pages prerendered, exit 0. (With the development env file the build compiles but fails prerendering `/_global-error`, because that file sets a non-production `NODE_ENV`; not related to this change.)
- `tsgo --noEmit`, biome, and `bun test src/test/headless-imports.test.ts` pass.

Process note: PR builds here run on Vercel only, and #38 merged with the preview build red. From now on a PR is merged only with the `Vercel – lexidraw` check green.


r4iju:
## Post-merge QA (master `d59d960d`, local dev build; the Vercel production deployment is SSO-protected so it was not smoke-tested from outside)

- The Vercel build for master succeeds again with the React Compiler on.
- A stored document opens in the editor with the inline-image plugin registered and no console or page errors from the app (the only console noise is the dev-only react-scan script blocked by CORS). Clicking a horizontal rule selects it and the `##` markdown shortcut produces a heading, so editor plugins still run.

Document editor on master after the drag-ghost fix

Issue #22, opened 2026-09-23
Split the light decorator nodes into the nodes package


r4iju:
## Parent

#13

## What to build

Image, inline image, video, YouTube, tweet, Figma, equation, page break, sticky, poll, and chart nodes move their serialization halves into the nodes package, obtaining their React component lazily so the package stays importable in Node. Blocks with no markdown form render as opaque placeholder comments carrying the node id.

## Acceptance criteria

- [ ] Documents containing any of these node types read as markdown; non-markdown blocks appear as id-carrying placeholders.
- [ ] Browser rendering and editing of these nodes unchanged.
- [ ] The nodes package still imports cleanly in a bun script.

## Blocked by

- #21

PR #41, opened 2026-09-23
feat: Headless markdown read for rich-text nodes

### LINKED ISSUE #21: Headless markdown read for rich-text nodes
## Parent

#13

## What to build

A new shared package holds the markdown transformers, the non-decorator node classes, and an exported registered node list, with no UI, framework, or CSS imports. The server runs a headless editor on top of it behind a `getMarkdown` procedure that returns the document as markdown with YAML frontmatter (id, title, updatedAt, tags, parent path), plus raw and Lexical-JSON variants. Documents containing node types the package does not yet register fail with a clear error listing those types.

## Acceptance criteria

- [ ] A document with headings, lists, code, tables, links, and quotes reads back as markdown with frontmatter through tRPC.
- [ ] Raw variant omits frontmatter; JSON variant returns Lexical state.
- [ ] A document with an unsupported node type returns an error naming the types.
- [ ] Browser editor behaviour unchanged; existing tests pass.

## Blocked by

- #18

r4iju:
Closes #21

## What

- New package `@packages/lexical-nodes`: the node classes that need no React component (autocomplete, emoji, keyword, mention, layout, collapsible), the markdown transformers, `html-to-text`, the emoji list, and the exported `CORE_NODES` / `CORE_TRANSFORMERS` lists. It depends on `lexical` and `@lexical/*` only. `CORE_NODES` is exactly the set the document editor registered before (emoji, keyword and mention are exported but were never registered, so they stay out).
- The browser editor composes its node list from `CORE_NODES` and its transformer list from `createTransformers([ARTICLE, IMAGE, EQUATION, TWEET])`, so a document is readable headlessly iff every node type it stores is in `CORE_NODES`.
- `documents.getMarkdown({ id, format: "markdown" | "raw" | "json" })` runs `@lexical/headless` on the server. `markdown` prepends YAML frontmatter (`id`, `title`, `path`, `updatedAt`, `tags`), `raw` is the bare markdown, `json` is the stored Lexical state. Unreadable ids and non-documents are `NOT_FOUND`; documents holding node types outside `CORE_NODES` fail with `UNPROCESSABLE_CONTENT` naming the types, and stored content that is not an editor state (`documents.create` accepts any JSON) fails with `UNPROCESSABLE_CONTENT` too. An editor state with no blocks reads as empty markdown.
- `entities.load` and the new procedure share one readable-entity query (`findReadableEntity`); `entities.getMetadata` and the frontmatter `path` share one ancestor walk (`entityAncestors`).
- `docs/agent-access.md` now describes the shipped `format` variants and the interim `UNPROCESSABLE_CONTENT` behaviour; placeholders for non-markdown blocks arrive with #22/#23.

## Notes for review

- **Horizontal rule.** The package uses the React-free `HorizontalRuleNode` from `@lexical/extension` (importing `@lexical/react` from a route handler crashes on the vendored RSC React). The app registers the React subclass right after `CORE_NODES`: both share the `horizontalrule` type, Lexical keeps the last class registered per type and instantiates it everywhere, so the selectable rule in the browser is unchanged.
- **Transformer order.** `createTransformers(extra)` puts the app extras before the table transformer, so `EQUATION` now precedes `EMOJI` (previously the reverse). Both are text-match transformers with disjoint patterns.
- **Table cells** keep their surrounding spaces on import (pre-existing importer behaviour); the tests compare trimmed cells.
- **Collapsible transformers** stay export-only, but the title is now emitted as a bold line. Before, the title's inline children were skipped by the block exporter and the title was silently dropped; this also changes the browser-side markdown export and the LLM tools that use the same transformers. Pinned by a package test.
- **`path`** is the slash-joined chain of ancestor directory titles plus the document title, bounded at 64 levels.
- **One lexical instance.** `lexical` declares an optional `typescript` peer, so the package pins the same `typescript` as the app; otherwise bun's isolated linker installs two `lexical` copies and node registration fails across them.
- **`agentRules: false`** in `next.config.ts` (own commit): `next dev` was writing `AGENTS.md`/`CLAUDE.md` into the app directory.
- **Bundle:** `@lexical/headless` brings `happy-dom` into the server bundle of the tRPC route. The local production build passes; worth a look at the function size after deploy.

## Verification

- `packages/lexical-nodes`: build, `bun test` (4 pass).
- `apps/lexidraw`: `tsgo --noEmit` clean, `bun test src/server/documents src/test` (7 pass), biome clean on changed files.
- Local production build mirroring Vercel (`NODE_ENV=production next build --turbopack`) succeeds.
- End-to-end against the worktree dev server with a `write` token (script output):

```
create rich: 200
create unsupported: 200
create drawing: 200
getMarkdown markdown status: 200
raw starts with frontmatter?: false
raw equals markdown body?: true
json root type: root
unsupported status/message: 422 Document contains node types without a markdown form yet: excalidraw, slide
create malformed: 200
malformed status/message: 422 Document content is not a Lexical editor state
create empty root: 200
empty root status/content: 200 ""
delete malformed: 200
delete empty: 200
drawing status: 404 Document not found
anonymous status: 404 Document not found
missing status: 404 Document not found
delete rich: 200
delete unsupported: 200
delete drawing: 200
deleted doc status: 404
```

  The `markdown` response for the rich document:

````
---
id: "qa-md-mudkn3yx"
title: "QA markdown doc"
path: "QA markdown doc"
updatedAt: "2026-09-23T03:56:52.918Z"
tags: []
---

# QA heading

Some **bold**, *italic*, ~~struck~~ and `code` text with a [link](https://example.com).

> A quote

- item one
- item two
    - nested

1. first
2. second

- [ ] todo
- [x] done

|  a  |  b  |
| --- | --- |
|  1  |  2  |

***

```ts
const x = 1;
```
````

- Browser check on the same server: editor renders a stored document with heading, rule, and table; clicking the rule selects it; the `##` markdown shortcut still produces a heading. QA evidence is posted as a comment after merge.

### CODE REVIEW

**Comment by r4iju** [comment]
## Post-merge QA (master `6da7d83d`, local dev build; the Vercel production deployment is SSO-protected so it was not smoke-tested from outside; its build for this commit succeeded)

HTTP against `/api/trpc` with a `write` token:

```
create rich: 200
create unsupported: 200
create drawing: 200
getMarkdown markdown status: 200
raw starts with frontmatter?: false
raw equals markdown body?: true
json root type: root
unsupported status/message: 422 Document contains node types without a markdown form yet: excalidraw, slide
create malformed: 200
malformed status/message: 422 Document content is not a Lexical editor state
create empty root: 200
empty root status/content: 200 ""
drawing status: 404 Document not found
anonymous status: 404 Document not found
missing status: 404 Document not found
deleted doc status: 404
```

Browser, signed in as the owner: a stored document with heading, formatted text, rule, table and an open collapsible renders; clicking the rule selects it (React `HorizontalRuleNode` still wins the `horizontalrule` registration); the `##` markdown shortcut produces a heading; no page errors. The same script gives identical results on the previous master commit.

Editor on merged master rendering a document with rule, table and collapsible

PR #44, opened 2026-09-23
feat: Append markdown to a document

### LINKED ISSUE #24: Append markdown to a document
## Parent

#13

## What to build

An `appendMarkdown` procedure converts markdown and appends it to the end of a document. An optional precondition on `updatedAt` rejects with a conflict error that carries the current `updatedAt`. Defines the conflict error type reused by insert and replace.

## Acceptance criteria

- [ ] Appended content appears at the end of the document in the browser.
- [ ] Existing blocks, including non-markdown blocks, are untouched.
- [ ] A stale precondition returns the conflict error with the current `updatedAt`; no precondition succeeds.

## Blocked by

- #21

r4iju:
Closes #24.

## What

- `documents.appendMarkdown({ id, markdown, ifUnmodifiedSince? })` converts the markdown with the headless editor and appends the resulting blocks to the stored state at the JSON level. Existing nodes are never re-parsed, so blocks without a markdown form survive byte for byte.
- The write is a compare-and-set on `updatedAt`. A save landing between the read and the write is retried against once when no precondition was given; with a precondition, or on a second loss, the call fails with `CONFLICT`.
- `StaleDocumentError` is the shared conflict type for append, insert, and replace. The tRPC error formatter surfaces it as `data.currentUpdatedAt` (ISO) so the caller can re-read and retry.
- `findWritableEntity`: owner, share with `EDIT` access, or public `EDIT`. Readers get `NOT_FOUND`, not a hint that the document exists.
- The append algorithm sits behind a small `DocumentStore` port with a drizzle adapter, which is what lets the compare-and-set and retry logic run under `bun test` without the environment-validating database package.

## Verification

Typecheck, app tests (20, including the append and conflict paths against an in-memory store), biome (33 diagnostics, same as master), and a production build pass locally.

Against a local dev server with an API token, on a document holding slide decks, drawings, a mermaid diagram, an article, a comment, and a thread:

| Call | Result |
| --- | --- |
| append without precondition | 200, 3 blocks, new `updatedAt` |
| append with the pre-append `updatedAt` | 409 `CONFLICT`, `data.currentUpdatedAt` equals the new `updatedAt` |
| append with the fresh `updatedAt` | 200 |
| empty markdown | 400 |
| malformed `ifUnmodifiedSince` | 400 |
| unknown document id | 404 |

The markdown read-back keeps every placeholder and ends with the appended heading, paragraph, list, and second paragraph, and the JSON state shows them as the last root children.

Browser QA is deferred to the consolidated pass after the feature lands.


r4iju:
Fresh-context review outcome (Opus subagent, verified against the code):

- **Fixed:** the compare-and-set token did not advance when two writes shared a millisecond, so one could clobber the other and a caller got back the `updatedAt` it sent. Every entity write (`documents.save`, `entities.save`, and the append store) now sets `max(updatedAt + 1, now)`; the invariant is stated once at `nextUpdatedAt`, and a same-millisecond test covers it.
- **Fixed:** whitespace-only markdown slipped past `min(1)` and appended a blank paragraph while the `EmptyMarkdownError` guard was unreachable. The input now refines on trimmed content and the dead guard is gone.
- **Fixed:** the drizzle store had no coverage; `document-store.test.ts` runs it against an in-memory libsql database (matching write, stale write, deleted row, monotonic token).
- Clean: write-access SQL (join constrained to the caller, unique share per user), the retry loop (bounded, re-parses on reread), block serialization for the browser, error mapping, and the test assertions.

Adjacent and pre-existing, not touched here: `documents.save` and `documents.delete` have no ownership check. Neither has a browser caller; tracked for a follow-up.

PR #46, opened 2026-09-23
feat: Insert markdown after a heading or at a block index

### LINKED ISSUE #25: Insert markdown after a heading or at a block index
## Parent

#13

## What to build

An `insertMarkdown` procedure inserts converted markdown after a heading matched by text, or at a block index as fallback. Ambiguous heading matches fail with the candidates listed; an `nth` argument disambiguates. The `updatedAt` precondition is mandatory.

## Acceptance criteria

- [ ] Insert after a unique heading places content directly below that heading's section start.
- [ ] Duplicate headings fail with a candidates list; `nth` selects one.
- [ ] Missing precondition is rejected as a usage error; stale precondition returns the conflict error.

## Blocked by

- #24

r4iju:
Closes #25.

## What

- `documents.insertMarkdown({ id, markdown, afterHeading | atBlockIndex, nth?, ifUnmodifiedSince })` places converted markdown directly below a top-level heading matched by text, or at a top-level block index. Exactly one target is required and the precondition is mandatory; both are usage errors (`BAD_REQUEST`) when violated.
- Heading matching folds whitespace and case, since callers type what they read off the rendered document. Duplicate matches fail with `data.candidates` (`nth`, `blockIndex`, `tag`, `text`) and the same list in the message; `nth` picks one.
- `atBlockIndex` counts top-level blocks from 0; `length` means append. Out of range is a `BAD_REQUEST` naming the range.
- Append is now the `end` placement of the same write path (`write.ts`), so both share the compare-and-set, the `CONFLICT` shape, and the router's error mapping. The placement is resolved on every attempt, so a retry after a lost compare-and-set follows what the other writer left.

## Verification

Typecheck, app tests (39, including placement resolution and the in-memory store paths), biome (33 diagnostics, same as master), and a production build pass locally.

Against a local dev server with an API token, on a document holding the heavy decorator nodes plus seeded headings `Intro`, `Notes`, `Notes`, `Outro`:

| Call | Result |
| --- | --- |
| no precondition | 400 |
| stale precondition | 409 `CONFLICT` with the current `updatedAt` |
| after unique `Intro` | 200, block lands directly under the heading |
| after `Notes` | 400 with two candidates (`nth` 1 and 2, blocks 12 and 14) |
| after `Notes` with `nth: 2` | 200, block lands under the second `Notes` |
| unknown heading | 400 |
| `atBlockIndex: 1` | 200, becomes root child 1; heavy nodes untouched |
| `atBlockIndex: 999` | 400 |
| both targets | 400 |

Browser QA is deferred to the consolidated pass after the feature lands.


r4iju:
Fresh-context review outcome (Opus subagent, verified against the code):

- **Fixed:** the docs and a comment promised that a retried write re-resolves the heading, but the mandatory precondition makes a lost compare-and-set an immediate `CONFLICT`, so no insert can retry. The claim is gone; the resolution stays inside the loop for the append path, and the test now says what it exercises.
- **Fixed:** heading text ignored `linebreak` nodes, so a two-line heading only matched its concatenation. Line breaks now fold as whitespace.
- **Fixed:** `nth` past the last match reused the "pass nth" message; it now says the `nth` is past the matches and lists them, with correct plurals.
- **Fixed:** a blank `afterHeading` was accepted and matched empty headings; it is now a usage error.
- **Fixed:** a missed heading now names the top-level headings that exist, and the docs say inline markup is ignored when matching.
- Clean: the zod usage rules surface as `BAD_REQUEST`, only root-level headings are candidates, the loop is bounded and non-mutating, append behaves exactly as on master, and the error formatter only adds `candidates`.

PR #47, opened 2026-09-23
feat: Serve entity load over REST with a public OpenAPI document

### LINKED ISSUE #27: OpenAPI tracer bullet: one REST endpoint and the public schema
## Parent

#14

## What to build

Wire the OpenAPI generator to the router so annotated procedures are served as plain JSON under `/api/v1`, starting with entity load. Serve the generated document unauthenticated at `/api/v1/openapi.json` with the bearer security scheme declared. Token auth applies on this path.

## Acceptance criteria

- [ ] curl with a bearer token fetches one entity as plain JSON from `/api/v1`.
- [ ] `/api/v1/openapi.json` is reachable without auth, validates, and lists the endpoint with the bearer scheme.
- [ ] Missing or revoked token returns 401 on the REST path.

## Blocked by

- #19

r4iju:
Closes #27.

## What

- `trpc-to-openapi` is wired to the router. Procedures that carry `meta.openapi` are served as plain JSON under `/api/v1` (no superjson envelope), starting with `entities.load` at `GET /api/v1/entities/{id}`, which now declares an honest output schema.
- The generated document is public at `/api/v1/openapi.json` with the `bearerAuth` scheme declared and a five-minute cache header.
- The REST transport exists for agents, so `createRestContext` requires a personal access token and answers 401 without one, then delegates to the shared `createTRPCContext` (which already rejects unknown or revoked tokens with 401). Auth itself is unchanged.
- A hermetic test checks the document's version, the entity path, and the bearer security scheme. The document is generated in a child `bun` process with `server-only` and the env stubbed via `--preload`, so the stubs never leak into the rest of the suite.
- `entities.load` now reports the sharer's user id in `sharedWith[].userId`; it used to be the share row's id (nothing in the app read it, but REST would have frozen it into the contract).
- The adapter is created per request, which is the library's signature. Outside production it regenerates the document on every REST call to validate the router; production skips that. The public document at `/api/v1/openapi.json` is prerendered at build time.

## Verification

Typecheck, app tests (44), biome (33 diagnostics, same as master), and a production build pass locally. The new routes appear as `/api/v1/[...trpc]` and `/api/v1/openapi.json`.

Live, against a local dev server:

| Request | Result |
| --- | --- |
| `GET /api/v1/openapi.json`, no auth | 200, OpenAPI 3.1.0, lists `/entities/{id}` with `security: [{ bearerAuth: [] }]`; validates with `@redocly/cli lint` |
| `GET /api/v1/entities/{id}` with a token | 200, flat JSON body |
| same without a header | 401 |
| same with `Bearer lxd_bogus` | 401 |

Lockfile note: `trpc-to-openapi` pulls `co-body`, which pins older `raw-body`, `type-is`, and `media-typer` at the top level; the modern versions moved to scoped keys, so no existing consumer's resolution changed.


r4iju:
Review outcome (fresh-context review, adjudicated against the code):

- **Accepted, fixed in 2be9eebb**: `mock.module` in `openapi.test.ts` is process-global under `bun test` and leaked the env stub into later test files. The document is now generated by a spawned `bun --preload` child; verified by a throwaway probe that the real `@packages/env` validation and the real `server-only` guard fire again in the suite process.
- **Accepted, fixed in 2be9eebb**: `sharedWith[].userId` was the `SharedEntities` row id because the map callback shadowed `entity`. No app code read the field (all consumers only count the array), so no UI change.
- Rebased onto master after #46; the error formatter now carries both `currentUpdatedAt` and `candidates`.

Typecheck, 44 app tests, biome (33, same as master), and a production build pass.

PR #48, opened 2026-09-23
feat: Replace a document from markdown with placeholder round trip

### LINKED ISSUE #26: Replace a document from markdown with placeholder round trip
## Parent

#13

## What to build

A `replaceMarkdown` procedure rewrites the whole document from markdown. Placeholder comments that still appear in the input re-insert the original node from the stored document; placeholders the caller removed delete those nodes. The `updatedAt` precondition is mandatory.

## Acceptance criteria

- [ ] get → replace with the same markdown leaves every block, including slides and Excalidraw embeds, intact.
- [ ] Removing a placeholder from the markdown removes that block.
- [ ] Missing or stale precondition is rejected.

## Blocked by

- #22
- #23
- #25

r4iju:
Closes #26.

## What

- `documents.replaceMarkdown({ id, markdown, ifUnmodifiedSince })` rewrites the whole document from markdown. The precondition is mandatory; a stale one is `CONFLICT` with `data.currentUpdatedAt`, and the write is the same compare-and-set as insert.
- Placeholder comments (`<!-- lexidraw:TYPE#N summary -->`) that still appear re-insert the original node from the stored document, byte-identical. A placeholder the caller dropped deletes that node. Summaries are never parsed. A block placeholder used inside a sentence, an unknown `TYPE#N` (the error lists what the document has), or the same placeholder twice is `BAD_REQUEST`.
- Inline placeholders (excalidraw, mermaid, comment, thread inside a paragraph) come back inline: the text node is split around the placeholder and keeps its format and style.
- Articles now export a placeholder line ahead of their derived prose so they survive the trip. On replace the prose blocks that still match are dropped with the article; edited ones stay as real content. The in-editor markdown view shows the same line, consistent with every other placeholder node.
- Response: `{ id, updatedAt, blocks, restoredPlaceholders, removedPlaceholders }`.

## Verification

Typecheck, package tests (16), app tests (59), biome (33 diagnostics, same as master), and a production build pass locally.

Live against a local dev server with a document holding two slide decks, two Excalidraw embeds, a Mermaid diagram, an article, a comment and a thread (block-level and inline):

| Case | Result |
| --- | --- |
| replace without `ifUnmodifiedSince` | 400 |
| stale precondition | 409, `data.currentUpdatedAt` set |
| replace with the markdown just read | 200, markdown re-read unchanged, root child types unchanged, every block-level and inline heavy node byte-identical |
| drop `excalidraw#2` and edit a paragraph | 200, that node gone, `excalidraw#1`, slides, article and thread kept, edit applied |
| `mermaid#7` | 400 naming the placeholders the document has |
| replace with a single line | 200, one paragraph remains |


r4iju:
Review outcome (fresh-context review, adjudicated against the code), all fixed in fe5fafb2:

- **Unsupported stored node types were silently erased** by a replace even though the read rejects them. Replace now runs the same guard and answers `UNPROCESSABLE_CONTENT`.
- **Placeholder-shaped text inside code** was resolved (or raised). Code blocks and inline code are literal now; that is also the documented escape hatch.
- **A summary containing markdown-active characters** (inline image inside a link: `logo`) overlapped the image text-match and deleted the node. Summaries now drop `[]()!*_`~$<>|`.
- **Article prose drop was context-dependent** (derived bullets merged into a following user list and were copied on every cycle; a code fence in the body swallowed the rest). The prose is now matched as the exact exported text before parsing; only an unchanged whole is dropped.
- **Tables grew one space per cycle** through the table import; cells are trimmed on import, with a fixed-point test in the package.
- **A block placeholder inside a table cell** was accepted into the cell; block placeholders are now top-level only, with matching docs wording.
- Small: `entityId ?? ""` in the article export, trailing tab tolerated, precondition checked before parsing, docs on mangled placeholders.

Package tests 18, app tests 70, biome 33 (same as master), production build pass; both QA scripts pass against the branch.

Issue #29, opened 2026-09-23
REST paths for the markdown procedures


r4iju:
## Parent

#14

## What to build

Expose get, append, insert, and replace markdown over `/api/v1` with the precondition and conflict error represented in the OpenAPI document.

## Acceptance criteria

- [ ] curl with a token reads a document as markdown with frontmatter from `/api/v1`.
- [ ] Append, insert, and replace work over REST; conflict responses carry the current `updatedAt`.
- [ ] All four appear in the OpenAPI document with their error codes.

## Blocked by

- #27
- #26

PR #49, opened 2026-09-23
feat: lexidraw CLI skeleton with auth, raw api, and schema commands

r4iju:
Closes #30.

## What

- New workspace `apps/cli`, binary `lexidraw`, compiled with `bun build --compile`. Root script `bun run cli:install` builds it and installs it to `~/.ai/bin/lexidraw`.
- Profiles `prod` (default, `https://lexidraw.app`) and `dev` (`http://localhost:3025`), chosen with `--profile` or `LEXIDRAW_PROFILE`; `LEXIDRAW_URL` overrides the base URL. Token from `LEXIDRAW_TOKEN`, else the macOS keychain (service `cli/lexidraw`, account = profile).
- `auth login [--token]` validates the token against the server and stores it in the keychain; `auth status` reports profile, base URL, token source, user, and scope.
- `api <METHOD> <path> [--json ...] [--query k=v]` is the raw escape hatch to `/api/v1`. Server errors are re-emitted as `{ code, message, status, issues? }` on stderr with exit 1; network and non-JSON failures get `NETWORK` and `BAD_RESPONSE`; usage errors exit 2 with `USAGE`.
- `schema <command>` prints method, path, operationId, parameters, and request body for a registered command from the OpenAPI document, cached per profile under `~/.cache/lexidraw/<profile>/openapi.json` for five minutes (`--refresh` bypasses). The registry starts with `auth status` and `doc get`; later tickets add the rest.
- Server: `GET /api/v1/me` (`auth.me`) returns user id, email, auth kind, and token scope, so the CLI can validate a token without touching an entity.
- On Darwin the build script re-signs the compiled binary: `bun build --compile` appends its payload after the linker signs the Mach-O, and macOS 27 kills the unsigned result on exec.

## Verification

CLI tests (60), app typecheck, app tests (45), biome (33 diagnostics, same as master), and a production build pass locally.

Live, from `/tmp` against a local dev server with `LEXIDRAW_PROFILE=dev` and no `LEXIDRAW_TOKEN`:

| Command | Result |
| --- | --- |
| `lexidraw auth login --token ...` | `{ profile, userId, email, scope: "write" }`, stored in keychain |
| `lexidraw auth status` | `tokenSource: "keychain"`, user and scope shown |
| `lexidraw api GET /entities/{id}` | entity JSON with title and elements |
| `lexidraw schema "doc get"` | `GET /entities/{id}`, path parameter `id`, cache file written |
[...]

Issue #45, opened 2026-09-23
documents.save and documents.delete have no ownership check


r4iju:
## What is wrong

`documents.save` and `documents.delete` in the documents router update or soft-delete any entity by id for any signed-in user (or any write-scoped API token). Neither procedure checks that the caller owns the document, has an `EDIT` share, or that the document is publicly editable.

Neither has a browser caller today (the editor autosaves through `entities.save`), so exposure is limited to direct API use, but the REST layer (#27) will make both reachable by path.

## What to build

Both procedures resolve the target through `findWritableEntity` (added in #44) and answer `NOT_FOUND` when the caller may not write it, matching `appendMarkdown`. A regression test covers the rejected case if a harness exists by then.

## Found by

Fresh-context review of #44.

PR #50, opened 2026-09-23
feat: REST paths for the entity subset

### LINKED ISSUE #28: REST paths for the entity subset
## Parent

#14

## What to build

Annotate entities list, create, save, update, delete, search, tags, share, and directory listing with REST paths. Flatten GET inputs where the generator requires it. Define the stable error code vocabulary the CLI will mirror.

## Acceptance criteria

- [ ] Each listed operation works over `/api/v1` with a token and appears in the OpenAPI document.
- [ ] Admin, TTS, backups, snapshot, image generation, and LLM procedures are absent from the document.
- [ ] Errors carry a stable `code` documented in the schema.

## Blocked by

- #27

r4iju:
Closes #28.

## What

- Thirteen entity procedures carry `meta.openapi` and an honest `.output()` schema, so they are served under `/api/v1` and appear in the OpenAPI document: `GET/POST /entities`, `GET /entities/search`, `GET/PUT/PATCH/DELETE /entities/{id}`, `GET/PUT /entities/{id}/tags`, `GET /tags`, `GET/POST /entities/{id}/shares`, `PATCH/DELETE /entities/{id}/shares/{userId}`. Directory listing is `GET /entities?parentId=...`.
- The document declares one `ErrorResponse` component (`{ message, code, issues? }`) whose `code` enum is the stable vocabulary in `error-codes.ts`; a test pins it to trpc-to-openapi's status mapping.
- Admin, TTS, backups, snapshot, image, and LLM procedures have no paths; the test asserts the document contains exactly the expected operations.
- `search` is registered before `load` because the adapter matches paths in registration order.

## Contract changes worth reading

- Path parameters are `id` everywhere. `drawingId` (shares) and `entityId` (tags) inputs were renamed, and the two UI callers updated; `getSharedInfo` now returns `entityId`.
- `update` awaits its UPDATE instead of firing and forgetting, so a 200 means the write landed.
- `create` returns the created summary, or `null` when the id is already taken (the `?new=true` drawing route re-runs create on refresh, so this is not a 409).
- `save` returns `{ id, updatedAt }`.
- `includeArchived` / `onlyFavorites` accept `boolean | "true" | "false"` instead of `z.coerce.boolean()`, which turned `?includeArchived=false` into `true`.
- `tagNames` / `entityTypes` accept an array or a comma-separated string; `entityTypes` gained `directory`; `parentId` is optional.
- `getEntityTags` returns `string[]` (the UI only read `.name`).

## Verification

Typecheck, app tests (67), biome (33 diagnostics, same as master), and a production build pass locally.

Live against a local dev server with a token, `/tmp/qa-rest-28.ts`: 46 of 46 checks pass, covering create → list → move into a directory → directory listing → filter → update → save → load → tags put/get/user tags/tag filter → search (not shadowed by `/{id}`) → share info → share unknown user 404 → share → change access → unshare → no token 401 → unknown id 404 → invalid body 400 with issues → delete → load deleted 404, and the document's operation set, forbidden-path absence, and error vocabulary.

Not covered: `CONFLICT`, `UNPROCESSABLE_CONTENT`, and `TOO_MANY_REQUESTS` are declared but only emitted by the markdown paths (#29) and rate limiting.


r4iju:
Review outcome (fresh-context review, verified live against the dev DB), all fixed in 159107c4:

- **Six published procedures performed no per-entity authorization** (`update`, `save`, `changeAccessLevel` never mentioned `input.id`; `delete` and `unShare` had no check; `updateEntityTags` clobbered other users' rows) and `getEntityTags` / `getSharedInfo` leaked other users' tags and collaborator emails. All pre-existing on master, but this PR was about to publish them as a token-facing contract. Every procedure now starts from `findReadableEntity` / `findWritableEntity` / `findOwnedEntity`; foreign ids are `NOT_FOUND`, `publicAccess` changes by a non-owner are `FORBIDDEN`. `list` tag joins are scoped to the caller too. The unbounded share-table scan and its PII log are gone. #45 (documents) becomes a two-line fix with the same helpers.
- **Contract**: `create` is idempotent for the owner and `CONFLICT` for anyone else's id (no more `200 null`); `save` on a missing id is 404 instead of a fabricated `updatedAt`; an omitted `appState` is left alone (only `null` clears it); `entityType`/`accessLevel` outputs are enums; `parentId` is an optional string; the comma-list branch documents its legal values; duplicate tag names are deduped and the response is re-read from the DB.
- **Error vocabulary** now includes the codes the adapter throws before any procedure runs (`PARSE_ERROR`, `UNSUPPORTED_MEDIA_TYPE`, ...), with a test that greps the adapter sources so a new one cannot go missing. `INTERNAL_SERVER_ERROR` bodies are masked (message and stack); route handlers log the original.
- Unit tests for the REST zod helpers; the QA script gained cross-user and shared-as-editor blocks (85/85 live).
- Rebased onto master after #48 and #49; `GET /me` joined the pinned operation set.

Typecheck, app tests (111), biome (33, same as master), production build pass.

Issue #31, opened 2026-09-23
Document, directory, and search commands


r4iju:
## Parent

#15

## What to build

`doc list|get|create|append|insert|put|delete`, `dir list|create`, and `search`. Entities are addressed by id or `--path` resolved through directory titles; ambiguous matches on writes fail with candidates, reads pick the most recent with a stderr warning, `--nth` disambiguates. `--format json|md|table`, raw markdown to stdout for markdown commands, `--page-all` as NDJSON. `doc put` replaces only with `--replace`, and insert or replace without `--if-unmodified-since` is a usage error.

## Acceptance criteria

- [ ] `lexidraw doc get --path "Dir/Title"` prints markdown with frontmatter.
- [ ] `lexidraw doc append <id> --file notes.md` succeeds; `doc put <id> --replace` without the precondition fails with a usage error.
- [ ] `lexidraw doc list --page-all` streams NDJSON; `--format table` renders a table.

## Blocked by

- #30
- #28
- #29

PR #51, opened 2026-09-23
feat: REST paths for the markdown procedures

### LINKED ISSUE #29: REST paths for the markdown procedures

### LINKED ISSUE #45: documents.save and documents.delete have no ownership check

r4iju:
Closes #29. Closes #45.

## What

- The four markdown procedures are served under `/api/v1` and appear in the OpenAPI document with tag `documents`:

  | Method | Path | Procedure |
  | --- | --- | --- |
  | GET | `/documents/{id}/markdown?format=markdown\|raw\|json` | `documents.getMarkdown` |
  | POST | `/documents/{id}/markdown/append` | `documents.appendMarkdown` |
  | POST | `/documents/{id}/markdown/insert` | `documents.insertMarkdown` |
  | PUT | `/documents/{id}/markdown` | `documents.replaceMarkdown` |

- Each write path declares 409 (and 400/401/403/404/422/500) against the shared `ErrorResponse`, which now has an optional `data` object with `currentUpdatedAt` (ISO, on a stale precondition) and `candidates` (on an ambiguous heading). The candidate shape is tied to `HeadingCandidate` at the type level.
- `getMarkdown`'s output is a union: `content` is a string for `markdown`/`raw` and the editor state object for `json`.
- The insert input's transform renders correctly through the adapter, so the tRPC contract is untouched.
- `documents.save` and `documents.delete` now resolve through the writable/owned reach helpers and answer `NOT_FOUND` for foreign ids (#45).
- Docs: path table rows, a "Preconditions over REST" note, and the roadmap now lists only the drawing paths as planned.

## Verification

Typecheck, app tests (119), biome (33 diagnostics, same as master), and a production build pass locally.

Live against a local dev server with a token, `/tmp/qa-rest-29.ts`: 58 of 58 checks pass. Read with frontmatter, raw, and json; append; insert after a heading; insert without the precondition is 400 naming `ifUnmodifiedSince`; a stale precondition is 409 with `data.currentUpdatedAt`; an ambiguous heading is 400 with `data.candidates`; replace with the markdown just read re-reads byte-identical; replace stale is 409; an unsupported node type is 422 on read; no token is 401; the document declares the four operations with their 409s; foreign `documents.save` and `documents.delete` are 404 and leave the row untouched.


r4iju:
Review outcome (fresh-context review, verified live): nothing blocking. Fixed in 8c6aac86:

- The insert body did not describe its placement rules, so a request satisfying every `required` field could still be a 400. `afterHeading`, `nth`, and `atBlockIndex` now carry `description`s stating the exactly-one rule, the heading match, and the index semantics; these reach `lexidraw schema` since it surfaces field descriptions.
- The read output is a `discriminatedUnion` on `format`, so the document ties `format` to the type of `content`.
- Left as is: `ErrorResponse.data` stays optional (always present in practice, conservative declaration). The CLI command registry gains the markdown verbs in #31.

PR #52, opened 2026-09-23
feat: Document, directory, and search commands in the CLI

### LINKED ISSUE #31: Document, directory, and search commands

r4iju:
Closes #31.

## What

- `doc list|get|create|append|insert|put|delete`, `dir list|create`, and `search`, all over the `/api/v1` paths. Every command is registered so `lexidraw schema "doc append"` describes it.
- Addressing by id or `--path "Dir/Sub/Title"` resolved through directory titles (exact match first, then case-insensitive). Parent directories are `--dir <id>` or `--dir-path <path>`, never guessed from the shape of the value. Titles containing `/` are addressed by id; the printed `path` is a display label. Ambiguous matches: reads take the most recently updated and warn on stderr with the candidate ids; writes fail with `AMBIGUOUS_PATH` and the candidates as JSON; `--nth N` picks by recency. A miss reports the parent id and the segment that failed.
- Output: JSON by default, `--format table` (fixed-width, no dependency), `--page-all` as NDJSON, and raw markdown to stdout for `doc get` (`--format md|raw|json`).
- Preconditions: `--if-unmodified-since <iso|latest>`; `latest` reads the document first and uses its `updatedAt`. `doc put` requires `--replace` and the precondition, `doc insert` requires the precondition; both are usage errors (exit 2) when missing. A 409 surfaces the server's `data.currentUpdatedAt`, an ambiguous heading its `data.candidates`.
- `doc create --file|--text` creates and then replaces against the revision the create answered, so the document starts at the caller's markdown rather than after the empty paragraph a new document carries. Blank bodies are rejected locally; a failed replace after a successful create reports the `createdId`.
- `doc delete <id>` reads the entity first and refuses anything that is not a document. For that, `GET /entities/{id}` now includes `entityType` (additive).

## Verification

CLI tests (168, including an in-memory entity and markdown stub server with revisions, preconditions, heading ambiguity, and the 404 delete contract), CLI typecheck and compile, app typecheck and tests (153), biome (33 diagnostics, same as master) pass locally after the rebase onto #53. The only app change is `entityType` on the entity load output.

Live from `/tmp` with the compiled binary against a local dev server: `dir create`; `doc create --dir "QA Dir" --file`; `doc get --path` with and without frontmatter (body starts at the heading); `append`; `insert --after-heading --if-unmodified-since latest`; `put` without `--replace` and without the precondition are usage errors; stale precondition is 409 with `currentUpdatedAt`; `put --replace --if-unmodified-since latest` succeeds; a second document with the same title makes `doc get --path` warn and `doc append --path` fail with candidates, `--nth 2` succeeds; `doc list --dir --format table`, `doc list --page-all`, `dir list --format table`, `search`; cleanup.

`--page-all` streams one row per line but does not page: `entities.list` returns a directory in one array with no cursor.


r4iju:
Fresh-context review outcome (adjudicated; fixed in a3d26b86, rebased onto #53 in c105cc0d):

**Blocking**
1. `doc delete <id>` deleted any entity type, including directories → the id form now reads the entity first and refuses anything but a document with `NOT_FOUND`. This needed `entityType` on `GET /entities/{id}`, which is now part of the load output (additive).
2. `--dir` decided id-vs-path with a UUID regex, but entity ids are free text → the form is explicit: `--dir <id>` is always an id, `--dir-path <path>` is always a path, and passing both is a `USAGE` error. The false "an id is a UUID" sentence is gone from the docs.

**High**
3. `doc create --file` orphaned an empty document when the replace failed → blank bodies are rejected locally as `USAGE` before any call, and a failed write after a successful create reports `createdId` in the error.
4. The ambiguity hint named `--nth` for intermediate segments where it cannot apply → hints now match the verb and segment (address the directory or document by id; `--nth` only for the final `--path` segment on verbs that take it).

**Medium**
5. Titles containing `/` are unaddressable by `--path`, and the printed `path` frontmatter is not safe to feed back → documented as a display label; such titles are addressed by id. No escaping added.
6. The test stub returned 200 on DELETE for any id → stub now 404s unknown ids and serves `GET /entities/{id}` with `entityType`.

**Low**
- `--page-all` now rejects any explicit `--format`, not just `table`.
- `zodError` is stripped from forwarded error `data` (it duplicated `issues`).
- Docs list exactly which commands take `--page-all` (`doc list`, `dir list`).

Skipped by decision: wide-glyph alignment in table output (cosmetic), and `--nth` out of range staying exit 1 `NOT_FOUND` (it carries `candidates`).

PR #53, opened 2026-09-23
feat: Drawing write with skeleton normalization

### LINKED ISSUE #33: Drawing write with skeleton normalization
## Parent

#16

## What to build

A drawing write procedure accepts either raw Excalidraw elements or the shorthand used by the official Excalidraw MCP (labels on shapes and arrows, bindings by id, frames with children, sticky notes). The server normalizes shorthand into canonical elements with the official skeleton converter under a DOM shim and a text-metrics provider, validates, and stores. Exposed over REST and as `drawing get|put|create` in the CLI. The skill gains the upstream element cheat sheet. Mermaid input is rejected with a clear error.

## Acceptance criteria

- [ ] A shorthand payload with two labelled boxes and a bound arrow stores as canonical elements with container ids, bound elements, and bindings, and opens correctly in the browser.
- [ ] Raw element payloads store unchanged apart from validation.
- [ ] `lexidraw drawing put` works from the CLI; the skill documents the format.

## Blocked by

- #27
- #30

r4iju:
Closes #33.

## What

Drawing write over REST and the CLI. `PUT /drawings/{id}` and `POST /drawings` accept a mix of canonical Excalidraw elements and the shorthand the official Excalidraw MCP uses (labelled shapes and arrows, bindings by id, frames with children, sticky notes). The server converts shorthand with the upstream skeleton converter under a scoped DOM shim, runs Excalidraw's `restoreElements` over the final set, and stores the result. `GET /drawings/{id}` returns the stored elements. The CLI gains `drawing get|put|create`; `docs/drawing-format.md` documents the accepted format.

## Design notes

- `packages/excalidraw-converter` pre-bundles `convertToExcalidrawElements` and `restoreElements` with React and mermaid stubbed, so the Next.js route only traces one chunk. That chunk is still ~3.7 MB and Vercel traces it into every serverless function that imports the tRPC root (about 22). Known trade-off, tracked for a follow-up rather than fixed here.
- Validation is dispatch-first: each element is classified as shorthand, canonical, or mermaid before its own schema runs, so every rejection is a 400 with the element index in the issue path and nothing unvalidated reaches the converter. Mermaid in any form is rejected by name.
- Canonical elements are normalised by `restoreElements` rather than stored byte-for-byte, so the browser can always open what the API accepted. The ticket's "unchanged apart from validation" wording is relaxed to this; the docs say so.
- `ifUnmodifiedSince` is required on `PUT`, consistent with markdown replace. `POST` covers the no-revision case.
- `parentId` on `POST` must be a writable directory; anything else is `NOT_FOUND` with no database text in the response.
- Element count is capped at 10,000; the 4.5 MB request body limit is documented.

## Verification

- `tsgo`, app tests (153), CLI tests (102), biome at the pre-existing 33 diagnostics, production build (only the pre-existing `adminTokens.list` / `tokens.list` prerender errors).
- Live against a dev server: 32 REST checks covering the skeleton round trip (container ids, bound elements, bindings, `frameId`, sticky note), the canonical round trip, malformed `points`, missing precondition, stale precondition (409 with `currentUpdatedAt`), unknown and non-directory parents, element cap, and both mermaid forms.
- Browser: the stored drawing opens with the frame, both labelled boxes, the bound arrow with its label, and the sticky note.
- Compiled CLI: `drawing create`, `drawing put --if-unmodified-since latest`, stale ISO → `CONFLICT`, missing flag → `USAGE`, `schema "drawing put"`.

## Review

A fresh-context review found three blocking items (unvalidated payloads reaching the converter as 500s, raw payloads the browser could not open, and a SQL leak on an unknown `parentId`) plus smaller ones. All are fixed in the second commit.


r4iju:
Fresh-context review outcome (all items adjudicated, all accepted, fixed in c1de0dba):

**Blocking**
1. Caller payloads reached the converter unvalidated and surfaced as 500 → dispatch-first validation per element (shorthand / canonical / mermaid), 400 with the element index in the issue path.
2. A raw payload could store a drawing the browser cannot open → strong canonical schema plus `restoreElements` on the final set; a throw is a 400.
3. `POST /drawings` with an unknown `parentId` returned 500 and leaked the SQL statement → parent must be a writable directory, else `NOT_FOUND`; insert wrapped so no database text reaches the client.

**Other**
- `dom-shim.ts` depth counter incremented before the shim was built → set only after globals are installed.
- Converter loaded eagerly → thunk, awaited only after validation passes.
- `ifUnmodifiedSince` optional on `PUT` → required, 409 added to `errorResponses`; CLI `drawing put` requires `--if-unmodified-since <iso|latest>`.
- Malformed `start`/`end`/`label` silently dropped → rejected.
- No element cap → 10,000 (`maxItems`), body limit documented.
- CLI tests for 409 passthrough and a missing `--file`.
- Root `overrides` pins `@excalidraw/excalidraw` to 0.18.1.

Not fixed by design: the converter chunk is traced into every function importing the tRPC root (see PR body).

PR #54, opened 2026-09-23
feat: lexidraw skill with an idempotent installer

### LINKED ISSUE #32: Lexidraw skill and install script
## Parent

#15

## What to build

A skill in the repo documenting the CLI for agents: auth, profiles, commands, addressing rules, write preconditions, and error codes. An install script symlinks it into the user's skills directory and installs the binary.

## Acceptance criteria

- [ ] `bun run skills:install` symlinks the skill and places the binary; re-running is idempotent.
- [ ] An agent given only the skill can read a document and append to it without further instructions.

## Blocked by

- #31

r4iju:
Closes #32.

## What

- `skills/lexidraw/SKILL.md`: the agent-facing skill for the `lexidraw` CLI. When to use it, setup (install, `auth login`, profiles, env overrides, the keychain-only-for-own-origin rule), the read-then-write flow with a worked example, addressing (`<id>` vs `--path`, `--dir` vs `--dir-path`, `--nth`, the `/`-in-title rule), placeholders, output formats, `schema`/`api` escape hatches, an error table, and the drawing commands with the shorthand cheat sheet from `docs/drawing-format.md`.
- `scripts/install-skills.sh` behind `bun run skills:install`: builds and installs the binary, then symlinks `skills/lexidraw` into `~/.ai/skills/lexidraw` (`LEXIDRAW_SKILLS_DIR` overrides). Idempotent, and it refuses to replace a real directory at the target.
- `docs/agent-access.md`: the Skill bullet names the live path and the installer.

## Verification

- Every command and flag in the skill was checked against the compiled binary's `--help` and `schema --list`; the frontmatter example matches the server's `withFrontmatter` output.
- Installer run twice: first run links, second reports `(already linked)`; the non-symlink refusal path was exercised.
- Acceptance test: a fresh subagent given only the skill text, a token, and a task ("append a paragraph to the document titled X") addressed the document by `--path`, chose `doc append` over `insert`/`put`, and verified the result, on the first attempt, with no skill edits.
- Biome does not lint markdown in this repo; `bash -n` on the script, tsgo and the CLI tests pass unchanged.

## Note

The symlink installed during development points at the worktree; `bun run skills:install` is re-run from the main checkout after merge.


r4iju:
Fresh-context review outcome (no blockers; all fourteen accuracy items accepted and fixed in 74e6daba / 5328a192):

**High**
1. The documented `drawing get` → `drawing put` round trip failed because `get` returns `{ id, title, elements, appState, updatedAt }` while `put --file` wants a bare array → the skill shows the `jq '.elements'` round trip and carries `updatedAt` into the precondition.
2. `docs/drawing-format.md` was referenced by repo path, unreachable from `~/.ai/skills` → `skills/lexidraw/drawing-format.md` is a repo-relative symlink to the doc, so it resolves through the installed symlink; the skill points at the file next to it.

**Medium**
3. `auth login --token` and the stdin path were undocumented → documented.
4. "Insert under a heading" misread as end-of-section → says "immediately after the heading block", and the worked example says where the bullet lands.
5. `--nth` orders by recency on paths but by document order on `--after-heading` → both stated.
6. The installer would pin the global symlink to a worktree path → it refuses to run from a worktree unless `LEXIDRAW_ALLOW_WORKTREE=1`.
7. Setup pointed at `cli:install` instead of this PR's `skills:install` → fixed.

**Low**
8–11. Error table now covers `PAYLOAD_TOO_LARGE`, `TOO_MANY_REQUESTS`, `NETWORK`, `BAD_RESPONSE`, `INTERNAL*`; `BAD_REQUEST` covers unmatched headings and out-of-range blocks (no `candidates` then); `FORBIDDEN` covers share rights; `--page-all` is listed per command and stated as exclusive with `--format`.
12. Shorthand cheat sheet corrected: `roundness` is `{ "type": 3 }`, frames take no geometry, `x`/`y` required, `fontFamily` dropped.
13. `dir list --nth` and the document-only nature of `doc list` are stated.
14. "Output is JSON" qualified by the formats.

The reviewer's own acceptance run (a fresh agent with only the skill text) succeeded in four commands with no failures.

Issue #36, opened 2026-09-23
Live drawing preview widget in MCP


r4iju:
## Parent

#17

## What to build

Register a preview widget as an MCP Apps resource so a drawing created through MCP renders live in the chat. Reuse the official Excalidraw MCP widget with three adapter tools mapping its checkpoint read, save, and export onto entity load and save, or build a thinner widget on the SVG export if the upstream repo's missing LICENSE file rules out reuse. Record the decision in the design doc.

## Acceptance criteria

- [ ] Creating a drawing through MCP shows it rendered in the chat client.
- [ ] Edits made in the widget persist to the entity.
- [ ] The license decision is recorded.

## Blocked by

- #35

PR #55, opened 2026-09-23
feat: stateless MCP endpoint over the same router and tokens

### LINKED ISSUE #35: MCP endpoint as a thin client of the API
## Parent

#17

## What to build

A stateless streamable HTTP MCP server hosted in the app, authenticated with the same bearer tokens, exposing tools that call the same procedures the CLI uses: list, read markdown, append, insert, and drawing put.

## Acceptance criteria

- [ ] Claude Desktop connected with a token lists documents, reads one as markdown, and appends to it.
- [ ] A revoked token is rejected on the MCP endpoint.
- [ ] No tool bypasses the router's permission checks.

## Blocked by

- #19
- #29
- #33

r4iju:
Closes #35.

## What

A stateless streamable-HTTP MCP endpoint at `POST /api/mcp`, hosted in the app, authenticated with the same personal access tokens as REST, exposing one tool per procedure the CLI already uses.

- `apps/lexidraw/src/app/api/mcp/route.ts`: resolves the token with the same `createRestContext` REST uses, answers 401 with the REST error body before any MCP machinery runs, then serves one message through `mcp-handler` (stateless mode, no sessions, no SSE fallback) with a server built around `appRouter.createCaller(ctx)`. Every tool call goes through the router, so ownership, share, and scope checks are the router's own.
- `apps/lexidraw/src/server/mcp/tools.ts`: `whoami`, `list_entities`, `search_entities`, `create_document`, `get_document_markdown`, `append_markdown`, `insert_markdown`, `replace_markdown`, `get_drawing`, `put_drawing`, `create_drawing`. Drawing tools reuse the router's exported element schema; the rest mirror the procedure inputs. Descriptions and the server instructions state the precondition rules and that the CLI's `latest` has no server-side meaning. This module is the seam where #36 registers the preview resource.
- `apps/lexidraw/src/server/api/error-body.ts`: the `/api/v1` error body and the `currentUpdatedAt` / `candidates` derivation, now shared by the tRPC error formatter and the MCP tool errors, so a `CONFLICT` or an ambiguous heading looks the same on both transports.
- `apps/lexidraw/src/test/server-runtime.ts`: lets a route's module graph load under `bun test` with an in-memory libsql database generated from the drizzle schema (the stored migrations start mid-history and cannot be replayed from empty).
- `docs/agent-access.md`: the MCP section documents the endpoint, auth, the tool list, and a `claude mcp add` snippet; #36 is the remaining item.
- After review: a 1 MB ceiling on the two read tools (refused whole with `PAYLOAD_TOO_LARGE`, never truncated), validation errors in the same body as REST, and a shared `resolveParentDirectory` so `entities.create` checks its parent the way `drawings.create` does.

Pinned: `mcp-handler@2.2.0` and `@modelcontextprotocol/server@2.0.0` (the peer mcp-handler 2.x requires). Lockfile side effect: root `chalk` hoists to 5.6.2; the packages depending on 4.x keep it nested.

## Verification

- `tsgo`, app tests (171, including 18 route tests: initialize, tools/list, read and write tool calls, missing token, revoked token, read-scope token on a mutation), biome at the pre-existing 33 diagnostics, production build (`/api/mcp` listed; only the pre-existing `adminTokens.list` / `tokens.list` prerender errors).
- Live with the MCP Inspector CLI over streamable HTTP against a dev server: `tools/list` shows 11 tools; `whoami`, `create_document`, `append_markdown`, `get_document_markdown`, `create_drawing`, `put_drawing` succeed; no token and a garbage token are 401; a read-scope token minted and then revoked through the app's token procedures is 401; a live read-scope token reads but gets `FORBIDDEN` on `append_markdown`; a stale precondition is `CONFLICT` with `data.currentUpdatedAt`.

## Not verified

Claude Desktop itself was not driven; the Inspector CLI is a real MCP client over the same transport and is the substitute. On a 401 the Inspector attempts OAuth discovery, which this endpoint deliberately does not publish; clients send the bearer header themselves.


r4iju:
Fresh-context review outcome (no blockers; all items accepted and fixed in 4bed963e):

**High**
1. Validation errors over MCP did not match the `/api/v1` body the docs promise (raw ZodError as `message`, message-only issues, no `data`) → `apiErrorBody` now produces the REST shape: `"Input validation failed"`, full `issues`, and `data` always carrying `currentUpdatedAt` and `candidates` (null when inapplicable). Both formatters derive those two fields from one function.

**Medium**
2. `get_document_markdown` could return ~10 MB into the model context → a 1 MB ceiling on `get_document_markdown` and `get_drawing`; over it the tool answers `isError` with `PAYLOAD_TOO_LARGE`, the actual size, and the REST path / CLI to use. Refused whole, never truncated.
3. Transport-level failures were swallowed → `onEvent` logs error events the way the REST route's `onError` does.

**Low**
4. `maxSubscriptions: 0`, so "no SSE" is literal.
5. Only `POST` is exported; GET/DELETE are Next's own 405 before auth.
6. `entities.create` inserted `parentId` unchecked (pre-existing, also reachable over REST and the browser) → a shared `resolveParentDirectory` used by both `entities.create` and `drawings.create`; a parent the caller cannot write to, or that is not a directory, is `NOT_FOUND`.
7. Tool schemas reuse the router's `MarkdownBody` and `AfterHeading`. Consequence, accepted and documented: arguments the tool's own schema refuses come back as the MCP SDK's plain-text protocol error (there is no hook to reshape it); router-side refusals keep the API body. The published JSON Schema is unchanged.
8. The config snippet says the 401 carries no `WWW-Authenticate`, so clients send the bearer header themselves.
9. Tests for the drawing tool schemas, cookie-only requests, and expired tokens.

PR #56, opened 2026-09-23
feat: Drawing render to SVG and PNG

### LINKED ISSUE #34: Drawing render to SVG and PNG
## Parent

#16

## What to build

A render procedure exports a drawing to SVG or PNG server-side using the pinned Excalidraw export helper under the DOM shim, with bundled fonts for PNG. Exposed over REST and as `lexidraw drawing render --format svg|png`.

## Acceptance criteria

- [ ] `lexidraw drawing render <id> --format svg` writes an SVG containing the drawing's text labels.
- [ ] PNG output renders text with the bundled fonts.
- [ ] The export package version is pinned and the call shape matches that version.

## Blocked by

- #33

r4iju:
Closes #34.

## What

Server-side drawing render to SVG and PNG, exposed as `drawings.render` (tRPC query), `GET /drawings/{id}/render?format=svg|png&scale=1..4` over REST, and `lexidraw drawing render <id> [--format svg|png] [--scale 1-4] [--out <file>]` in the CLI.

## Design notes

- **SVG** comes from `exportToSvg` in the pinned `@excalidraw/excalidraw` 0.18.1, re-exported from `packages/excalidraw-converter` (same bundle and stubs as the write path) and run under the DOM shim. The 0.18.1 call shape is written down as a type in the converter package, since the pinned package re-exports it untyped from `@excalidraw/utils`. Font inlining is skipped because it is the only part of the export that fetches over the network; the SVG names the families instead. With no I/O inside the export, the DOM shim's window stays scoped to one microtask drain, which is now the documented contract of `withDomShimAsync`.
- **PNG** rasterises that SVG with `@resvg/resvg-js` 2.6.2 (exact pin). resvg cannot read WOFF2, so `bun run fonts:sync` decompresses the editor's faces to TrueType into `apps/lexidraw/src/server/drawings/fonts/` (25 faces, 1.1 MB, checked in) and generates the load order: each family's Latin subset first, hand-drawn families first, because resvg answers a family with the first matching face and a missing glyph with the first face that has it. The CJK family (200 faces, 25 MB) is excluded and documented. `next.config.ts` externalises resvg and traces the font directory into the serverless functions.
- **Response shape** is JSON `{ id, format, contentType, encoding, width, height, data, updatedAt }` with `data` as SVG text or base64 PNG. The installed `trpc-to-openapi` 3.3.0 hardcodes JSON responses, and a second raw route would have duplicated auth and error mapping outside the generated contract. The CLI decodes it.
- Scale caps at 4 and output at 16 megapixels (`BAD_REQUEST` above); the encoded image caps at 3 MB (`PAYLOAD_TOO_LARGE`), so the JSON body stays under Vercel's response limit.
- After review: Excalidraw's family names are remapped to the bundled faces on the copy handed to resvg (`Helvetica` → `Liberation Sans` deliberately), deleted elements are excluded before export, and `fonts/LICENSES.md` ships with the faces.
- Font licences per upstream: Comic Shanns MIT; Nunito, Lilita, Liberation, Cascadia OFL 1.1; Virgil and Excalifont are Excalidraw's own, redistributed from the MIT package the app already renders with. Excalifont's name table carries an "All rights reserved" string; flagged, not treated as blocking.

## Verification

- `tsgo` in the app, CLI, and converter package; app tests (188, including the render and font-map tests: label text in the SVG, no embedded fonts, globals restored, PNG signature/size/scale, the pixel cap, and a pixel check that the label region is dark and the area below it is not; emptying the font list makes that check fail); CLI tests (174); compile; biome at the pre-existing 33; production build (only the pre-existing `adminTokens.list` / `tokens.list` prerender errors; the traces contain the 25 fonts and the resvg binding).
- Live against a dev server with the compiled CLI: SVG contains `Ingest`, `Index`, `batches`; the PNG at 2x renders the labels in Excalifont; bad format, scale 9, unknown id, and no token return 400/400/404/401.

## Caveats

- `import.meta.dir` is undefined under turbopack, so the font directory resolves from `process.cwd()` first. That holds for `next dev`, `next start`, and Vercel's function layout, but a real Vercel render was not exercised (prod is SSO-protected).
- The CLI fixture `apps/cli/test/fixtures/openapi.json` is regenerated from the server and now carries the full document.


r4iju:
Fresh-context review outcome (all items accepted, fixed in d3832938, rebased onto #55):

**Blocking**
1. Four of the eight families Excalidraw names in the SVG (`Helvetica`, `Cascadia`, `Nunito`, `Comic Shanns`) matched no bundled face's name-table family, so `defaultFontFamily` drew them in Excalifont → the sync script now parses each face's name table and emits an explicit family map; the render rewrites `font-family` only on the copy handed to resvg (`Helvetica` → `Liberation Sans` is the one deliberate, metric-compatible substitution). Font name tables are untouched, so no OFL reserved-name issue. Tests pin the map to the name tables and assert the remapped families render with a different ink width than Excalifont.

**High**
2. Deleted elements inflated the canvas bounds → filtered before export; the converter's export option type now carries the NonDeleted constraint and mirrors the 0.18.1 declaration.
3. The 16-megapixel cap did not bound the response (a 7.2 MB JSON body would exceed Vercel's limit) → a 3 MB ceiling on the encoded image, `PAYLOAD_TOO_LARGE` (413) with advice to lower `scale`, documented alongside the pixel cap.

**Medium**
4. No licence text shipped with the fonts → the sync script writes `fonts/LICENSES.md` with each family's licence, the name-table copyright, and the OFL 1.1 and MIT texts.

**Low**
5. CLI `bytes` now uses `Buffer.byteLength`. 6. Reported sizes are the image's own (PNG pixels, SVG unrounded units). 7. Docs state that only `viewBackgroundColor` is honoured. 8. Fonts are traced into every server entry, not two routes (1.1 MB per function). 9. Docs say image elements draw nothing while their bounds reserve the space.

PR #57, opened 2026-09-23
feat: live drawing preview widget in MCP

### LINKED ISSUE #36: Live drawing preview widget in MCP

r4iju:
Closes #36.

## What

A drawing preview widget registered as an MCP Apps resource, so a drawing created or edited through MCP renders live in a chat client and edits made in the widget persist to the entity.

- `packages/drawing-widget`: a new workspace package built with `Bun.build` at package build time (mirroring the converter package), not committed as an artifact. The entry bundles React, `@excalidraw/excalidraw`, and the ext-apps bridge into one inlined HTML document. Turbo's existing build edge means `next build` and Vercel need no extra step. Mermaid is stubbed out, matching `put_drawing` refusing it anyway.
- `apps/lexidraw/src/server/mcp/widget.ts`: registers `ui://lexidraw/drawing-preview` on the same stateless authenticated handler; `get_drawing`, `put_drawing`, and `create_drawing` carry `_meta.ui.resourceUri`. The HTML is imported lazily inside the resource read, so it never sits in the request path.
- What the widget renders travels in the result's `_meta`, and the shape follows what the tool's own answer already carries. `create_drawing` and `put_drawing` answer with an id and counts, so their payload carries the elements — read back after the write, so the widget shows what the server normalised rather than what was sent, and past 1 MB it says so instead. `get_drawing` answers with the whole drawing, so its payload is `{ id, title, updatedAt, canWrite }` and the widget takes the elements out of the answer: one drawing on the wire, under one ceiling.
- The payload is built for a client that advertises MCP Apps, and for one whose capabilities this endpoint never saw — which is the usual case here, because the endpoint is stateless and a `tools/call` arrives with no memory of the `initialize` that named them. A client that negotiates and does not ask for MCP Apps pays nothing: no scope lookup and no read-back of the drawing.
- The host owns every call the widget makes and can deny, drop, or time one out. A refused save keeps the edit and offers a retry instead of leaving the header mid-save; a refused load says so.
- A read-scope token yields `canWrite: false`, which puts the widget in Excalidraw's view mode with a read-only label.

## Licence decision (acceptance criterion 3)

`excalidraw/excalidraw-mcp` has no LICENSE file, so none of its code or built widget is reused. This widget is ours, built on the MIT-licensed `@excalidraw/excalidraw` package the app already renders with. Recorded in `docs/agent-access.md` and its research notes.

## CSP

One resource domain, `https://esm.sh`, no connect domains. The widget sets `EXCALIDRAW_ASSET_PATH` to the pinned version's dist there, and both halves read that origin from one shared module, so the declared domain cannot drift from the one the widget loads. No connect domain is needed: every call the widget makes is a tool call over the host bridge on the host's own authenticated session, and the widget holds no token.

The editor's chrome is cut to match rather than the policy widened to fit it: export, save-to-file, open, and the image tool are off through `UIOptions`, and the library button is hidden. Those reach a file system the sandbox has none of, or fetch subsetted fonts and a wasm encoder from origins the policy does not name, and images are binary files `put_drawing` does not carry.

## Size

`resources/read` answers about 5 MB, authenticated and uncached. The document is built once per instance — it is a string literal in a module, so the module cache is the memoisation — and the build fails if it grows past 6 MB, so the size stays a decision rather than a drift. The transport answers it as an event stream rather than one buffered body (`content-type: text/event-stream`, no `content-length`), so Vercel's buffered-response limit should not apply; that was checked against the route, not against a deployment.

## Verification

- `tsgo` clean, app tests (197, including the route tests for the tool `_meta` in both key forms, `resources/list`, `resources/read`, the payload holding what the server stored rather than what was sent, the read payload carrying no second copy of the elements, the too-large case, unauthenticated `resources/read` → 401, and unit tests for who the payload is built for), CLI tests (174), widget tests (7, in Chromium against the SDK's own host bridge: the save round trip with its precondition, a read's elements taken from its own answer, a refused save keeping the edit until a retry, a refused load reported rather than hung, the host's theme, and the absent export/save/library chrome), biome at the pre-existing 33, production build (only the pre-existing `adminTokens.list` / `tokens.list` prerender errors).
- Live against a dev server: `resources/list` with the right mime type and CSP, `resources/read` at 4.95 MB, `tools/list` carrying the meta on exactly the three drawing tools, a real `create_drawing` whose meta held the normalised elements, a `get_drawing` whose meta held four fields and no elements while its answer held them, a read-token connection reporting `canWrite: false`, and unauthenticated `resources/read` → 401. A host page rendered that drawing, saved an edit through `put_drawing`, and a follow-up read showed the stored result.

## Not verified

No shipping MCP Apps host was available here. Everything above was proved against the SDK's own host implementation and the live endpoint; whether a production chat client renders the resource, forwards `_meta`, and grants model-context updates is unconfirmed. The widget falls back for the middle case and feature-detects the last.

## Review follow-up

A fresh-context review found no blockers. Fixed in the second commit: a refused tool call used to strand the widget on "Saving…" and drop the edit, so both the save and load paths now catch protocol rejections, keep the pending edit, and offer a retry; the theme now comes from the host handshake rather than defaulting to light; the editor actions that cannot work in a sandbox (export, save, open, image, library) are turned off rather than widening the content security policy to admit them; `get_drawing` no longer ships the elements in both the content and the metadata, so a near-ceiling read stays one payload; the build fails if the document outgrows its size budget; and the generated constants module refuses to emit anything that is not data.

One judgement call worth naming: the capability gate that would skip the preview read for a plain client cannot fire here, because a stateless endpoint builds a fresh server per request and never sees the client's negotiated capabilities. Unknown therefore means yes, which is written at the definition and in the docs. A plain client pays one indexed read per drawing write; the gate starts saving it the moment capabilities are available per request.


r4iju:
Fresh-context review outcome (no blockers; all items accepted, fixed in 98a7b36d):

**High**
1. `save()` and `load()` caught tool errors but not protocol rejections (host denies the call, timeout, bridge closed), so a refused call left the widget on "Saving…" forever and silently dropped the edit → both paths catch, the edit stays pending, an error state names the failure and offers a retry, and a guard stops one load starting another. Tests drive a host that throws.

**Medium**
2. The widget always opened in light theme, because the initial state read the host context before the handshake delivered it → the theme comes from the resolved handshake, with the change listener kept. Tested against a dark host.
3. The editor still offered export, save to disk, open, image and the library, none of which can work in the sandbox, and making them work would have meant widening the content security policy → those actions are off and the library trigger is hidden. The policy is unchanged: one resource domain, no connect domains.
4. `get_drawing` shipped the elements in both the model-visible content and the metadata, so a read that was capped near 1 MB could put 2 MB on the wire → the read's metadata carries only id, title, `updatedAt` and `canWrite`; the widget takes the elements from the content it already has, and the bridge fallback remains for hosts that drop metadata. The write payloads still carry elements, which is where that choice earns its keep.
5. The re-read that builds a write preview is now gated on the client supporting MCP Apps. Worth naming: on a stateless endpoint the negotiated capabilities are never visible on a tool call, so unknown means yes, written at the definition and in the docs. A plain client pays one indexed read per drawing write; the gate starts saving it if capabilities ever arrive per request.
6. The 5 MB resource is unchanged in design (loading the editor from a CDN at runtime would put a third party in the widget's script path). The module cache is the memoisation, the build now fails past a 6 MB budget, and the docs record that the transport answers it chunked with no content length, verified against the route and not against a deployment.

**Low**
7. The generated constants module refuses to emit anything that is not data. 8. A dead font-URL rewrite is gone. 9. Size reporting was already on by default; the test now asserts it.

Issue #17, opened 2026-09-23
Agent access phase 6: MCP server with preview widget


r4iju:
Design: `docs/agent-access.md` (MCP section). Phase 6 of 6. Depends on phases 1, 3, and 5.

## Scope
- Host an MCP server inside the Next app with `mcp-handler`, stateless streamable HTTP, bearer token auth through the same context check. Tools are thin clients of the same procedures the CLI uses.
- Reuse the built preview widget from `excalidraw/excalidraw-mcp` as an MCP Apps resource, with three adapter tools mapping its checkpoint read/save/export onto entity load and save.
- Decision at implementation time: that repo's README says MIT but it has no LICENSE file. If that is unacceptable, build a thinner preview widget on the SVG export helper from phase 5 instead.

## Acceptance
- Claude Desktop connected with a `lxd_` token can list documents, read one as markdown, and create a drawing that renders live in the widget.
- A revoked token is rejected on the MCP endpoint.

r4iju:
## QA pass on the shipped feature

Run against a local dev server on the merged `master` (`cde2af60`), covering everything from #18 through #35. Prod is SSO-protected, so it could not be smoke-tested there.

| Area | Checks | Result |
| --- | --- | --- |
| Tokens, browser and auth boundary | 13 | Pass |
| REST `/api/v1` | ~60 | Pass, 3 contract nits |
| CLI | ~45 | Pass |
| MCP `/api/mcp` | 17 | Pass, 1 nit |
| Browser sanity after the server changes | 5 | 1 failure |

Verified end to end: minting and revoking tokens in the settings page, and a revoked token rejected by both REST and MCP; all 24 OpenAPI paths present and matching the design table; preconditions answering 409 with `currentUpdatedAt`, ambiguous headings answering 400 with candidates, unreadable node types answering 422; a read-scope token refused on every mutation; foreign and unknown ids answering not-found; shorthand expanding to canonical elements with labels, bindings, frame membership and ids preserved; placeholder semantics exactly as documented, with an unmodified markdown round trip lossless; the CLI's whole surface including exit codes and machine-readable errors; the skill's worked example running verbatim; the eleven MCP tools with their conflict and forbidden bodies and the read ceiling.

![Personal access tokens settings page with two tokens minted](https://lch58seqqbwhqunw.public.blob.vercel-storage.com/2026/09/lexidraw-qa/qa-final-tokens-Lk48lfMclMLX0YplSS1z8oWL9gwhiq.png)

[...]

PR #60, opened 2026-09-23
fix: QA follow-ups on agent access

r4iju:
Follow-ups from the QA pass on the agent-access feature. Evidence and the finding list are on #17.

## What

- **The cache-tag invariant is written down and pinned, but it does not invalidate anything yet.** A single module owns the pair: a tag per entity for the render, and one revalidation helper for the writers, so procedures stay callable from the REST route, the MCP route, and a render alike. Every cached page an API write can change tags itself (document, print, drawing, url, dashboard), and every writer of an entity row — not only the REST and MCP ones — drops the entity, the directory it is listed in, and the listing above that when the tree changed shape. What it does **not** do is make a stale browser fresh: all five pages are `"use cache: private"`, Next keeps private entries in the Resume Data Cache where `revalidateTag` cannot reach them, and the staleness a browser actually shows is the client router cache, which no server-side call can drop. The observable fix is a client refresh signal, filed as #61. This PR makes the invariant explicit and tested so it is load-bearing the day a page stops being private, rather than rediscovered from scratch.
- **The MCP tool schemas are portable again.** A drawing point is expressed as a two-element array rather than a tuple, so the generated JSON Schema no longer contains a bare `false` that strict clients reject. List-valued `type` (from nullables, which cannot be fixed at the source) is rewritten to `anyOf` at both publication boundaries, REST and MCP, by one shared helper that leaves data-valued keywords alone and never displaces an existing `anyOf`. `create_document` and `create_drawing` publish a plain optional `parentId` and still read `null` as "the root", which is what the router and the REST path take.
- **Creating a drawing reports `elementCount`,** like replacing one does. It is the call where the stored count is least predictable, because shorthand expands.
- **A wrong method on a live path answers 405 with `METHOD_NOT_SUPPORTED`** and an `Allow` header, instead of 404. The method table is derived from the generated OpenAPI document, so a new path cannot drift out of it; `HEAD` is served as the `GET` Next routes it through, and unpublished paths still answer 404. The refusal is answered before authentication on purpose — it says no more than the public OpenAPI document does.
- **Error messages name the right thing.** Loading or deleting a document no longer says "Drawing not found", replacing a drawing no longer says "Document was modified"; the precondition error now requires its subject at the type level, so a call site cannot pick the wrong noun silently. The unsupported-content-type message has its quotes closed.

## Verification

- `tsgo`, app tests (231), CLI tests (174), widget tests (7), biome at the pre-existing 33, production build with only the pre-existing `adminTokens.list` / `tokens.list` prerender errors.
- Live against a dev server: the MCP inspector in strict mode goes from 6 errors across 3 tools to 0 errors; `POST /api/v1/me` answers 405 with `Allow: GET`, `HEAD /api/v1/me` answers 204 as it did before, and an unknown path still answers 404; a wrong content type answers 415 with a balanced message; creating a drawing from two shorthand elements reports three stored, which is the label expanding into its own text element.

## What this does not fix

The QA reproduction was a browser showing pre-write content. Hard navigation and reload are fresh on both master and this branch; a soft navigation through a prefetched link stays stale on both, and stays stale even with the page's cache directive removed entirely, which places it in Next's client router cache. Server-side revalidation cannot reach that, and it cannot reach a private cache entry either (`updateTag` and `refresh` can, and both are Server-Action-only). Closing the observable staleness needs a client refresh signal: #61.


r4iju:
The client half of the stale-render finding is filed as #61: an open editor still misses an out-of-band write on a soft navigation, which is Next's client router cache and needs a refresh signal rather than server-side revalidation.

r4iju:
Review outcome — all eight findings accepted, fixed in a7e6eea.

**High — the tag pair is inert.** Accepted as stated. The A/B against master is decisive: the pages are `"use cache: private"`, Next keeps private entries in the Resume Data Cache, and `revalidateTag` never reaches them. I kept the pair rather than removing it — it is the correct pairing and now pinned by tests, so it is load-bearing the day a page stops being private — but the review was right that landing it described as a fix is not defensible. The module doc now opens by saying the tags do not currently invalidate anything and why, and the PR body's headline no longer claims a stale render is fixed. The observable fix stays #61.

**High — `HEAD` answered 405.** Real regression; Next serves HEAD through the exported GET. Mapped to GET before the method lookup: `HEAD /api/v1/me` is 204 again, `POST` is 405 with `Allow: GET`. Pinned by a test.

**Medium — `parentId: null` rejected.** Real regression. `parentId` is now `z.preprocess(v => v ?? undefined, z.string().optional())`, so null reads as the root as it did on master while the published schema stays a plain optional string — confirmed from a live `tools/list`.

**Medium — the bare swallow.** Narrowed to Next's `E7`/`E181`. Rather than leave every caller tolerating a missing store, the three workflow steps that legitimately run with no request call `revalidateEntitiesOutsideRequest`, so the exceptional context declares itself. This surfaced a real consequence: procedures driven outside a Next server were failing silently through the old swallow, which is why `installServerRuntime()` now installs a documented `next/cache` stub.

**Low — incomplete coverage.** Covered rather than narrowed, and wider than the list: the three share mutations revalidate the parent, tree-shape writes revalidate the grandparent through `revalidateEntitiesAndParents`, and the browser-only writers (`documents.create/save/delete`, `updateUserPrefs`, both thumbnail paths, distill, the snapshot procedures, the admin procedures, the workflow steps) are in. The module's "every writer of an entity row" claim is now true.

**Low — `type` + `anyOf`.** Guarded: the `anyOf` is kept and the type union goes under `allOf`. I chose that over throwing because the rewrite runs at module load, and a crash there takes the app down for a schema that is still expressible. The rewrite also no longer descends into `const`/`default`/`enum`/`example`/`examples`. Key order is preserved, so the generated document is byte-identical and the CLI fixture needed no regeneration.

**Low — 405 before auth.** Kept there deliberately and documented: moving it after auth would make the refusal depend on the caller, which the already-public OpenAPI document contradicts.

231 app tests (from 214), 174 CLI, 7 widget, tsgo clean, biome at the baseline 33, production build green with only the pre-existing prerender errors.

PR #62, opened 2026-09-24
fix(cli): point prod at the real deployment and verify the host before sending a token

r4iju:
## What

The CLI's `prod` profile defaulted to `https://lexidraw.app`, which is **not this app**. It is an unrelated Excalidraw fork hosted on Railway. Prod is the Vercel project at `https://lexidraw.vercel.app`. So `lexidraw auth login` and every command on the default profile sent the personal access token to a third party. No token was ever stored for `prod` on this machine, so nothing leaked, but anyone following SKILL.md would have hit this.

This PR fixes the default URL and closes off the whole class of problem: `requestApi` is the only place a token is attached. Before the first authenticated request, it awaits the session's one-time check that the base URL itself serves an OpenAPI document titled `Lexidraw API`. The check shares the 5-minute schema cache, and the CLI follows no redirects. If the host answers with another site, another API, an empty body or a redirect, the command fails with `NOT_LEXIDRAW_SERVER` and no token is sent. An outage (5xx) or an unreachable host keeps its own code. The check runs on the first request, so usage errors still exit 2 without touching the network.

Also fixed: `docs/agent-access.md`, SKILL.md, and the `claude mcp add` example now point at the real host. The reader's user-agent no longer names someone else's site. SKILL.md's frontmatter example now shows `path` without the leading `/` the server never emits, and its error table gains `NOT_LEXIDRAW_SERVER` and `REDIRECT`.

## Options considered

1. Change the constant only. This fixes today's bug, but a mistyped `LEXIDRAW_URL`, an `http://` URL, or a future domain change would hand the token to whoever answers.
2. **Also verify the host before any token leaves (this PR).** It costs one cached, unauthenticated GET (59 KB, about 140 ms) every 5 minutes per origin.

## How I verified it

- Tests first. There are 28 impostor tests: 6 token-sending commands plus `auth login`, each against four hosts (an HTML site, a JSON API with another title, an empty 200, and a 308 to a real Lexidraw stub). There is also a test that the redirect target is named, one that a 5xx is not called "wrong host", and five tests that argument errors exit `USAGE` against an unreachable host. All of these were seen failing first. `cd apps/cli && bun test`: 209 pass. `tsgo --noEmit` and `bun run compile` are clean. The Biome count stays at the 33 baseline.
- Real prod smoke test, default profile: `auth status`, `search`, `doc get/create/append/insert/put` (including the 409), tags, `drawing create/get/put/render` (svg+png), and MCP `initialize`/`tools/list`/`tools/call`/`resources/read`. All pass on `lexidraw.vercel.app`.
- `LEXIDRAW_URL=https://lexidraw.app` fails with `NOT_LEXIDRAW_SERVER`. `LEXIDRAW_URL=http://lexidraw.vercel.app` fails with `NOT_LEXIDRAW_SERVER` and names the https address. Neither sends a token.

After merge, rebuild the installed binary with `bun run skills:install`.


r4iju:
**Review outcome.** A fresh-context review found five correctness issues. I checked each against the code, and all five were real. They're fixed in 1bf5717f, with tests that were seen failing first:

1. **Redirects got past the check.** Fetch followed redirects, so a host that redirected `/openapi.json` to Lexidraw passed and then received the token. `http://lexidraw.vercel.app` sent it in cleartext before the 308. Fixed: `redirect: "manual"` everywhere, and a 3xx is `NOT_LEXIDRAW_SERVER` (or `REDIRECT` on an authenticated call).
2. **Usage errors hit the network first.** Opening the session eagerly turned `USAGE`/exit 2 into `NETWORK`/exit 1 in `doc`, `dir` and `api`. Fixed: verification moved into `requestApi`, so it runs on the first authenticated request, and opening a session is synchronous again.
3. **An outage read as the wrong host.** A 5xx with a non-JSON body was reported as `NOT_LEXIDRAW_SERVER`. It now keeps `BAD_RESPONSE` with its status.
4. **An empty 200 crashed** with `INTERNAL`. A document that isn't an object is now rejected before it is cached.
5. **The docs overclaimed.** Rewritten to match the behaviour.

Standards notes I took: the test-count claim in the body is corrected. Notes I left: `Target = string | ApiSession` stays, because the union is what makes an anonymous call visibly tokenless. The title comment on the server side stays, because it warns whoever renames the title, which is a separate reason from the one at `ApiSession`.

r4iju:
Second review pass, adjudicated and fixed in 2b892380:

- **An old-format cache could approve a redirecting host.** Accepted. Each cache entry now records the origin that served it as `{ origin, document }`. An entry with no origin, or with a different one, gets refetched. Tests cover both.
- **`originKey` collisions (`host:8080` vs `host-8080`).** Accepted, handled by the same origin check. A collision now causes a refetch instead of letting one host vouch for another. The key itself is unchanged, since it only needs to be directory-safe.
- **Redirect advice pointed at `/api/v1/openapi.json`.** Accepted. `NOT_LEXIDRAW_SERVER` after a redirect now says `Point LEXIDRAW_URL at <base>.`, with the document path removed. The SKILL `REDIRECT` row no longer says 'authenticated call' only.

Gate: 211 CLI tests pass, tsgo is clean, compile works, and Biome reports no new diagnostics.

r4iju:
QA on prod (lexidraw.vercel.app, merged build), CLI installed from master via `skills:install`:
- `lexidraw --profile prod auth status`, with no LEXIDRAW_URL set, reaches https://lexidraw.vercel.app and returns the user and scope. Exit 0.
- `LEXIDRAW_URL=https://example.com` gives `NOT_LEXIDRAW_SERVER` ('does not serve the Lexidraw API; no token was sent'). Exit 1.
- `LEXIDRAW_URL=http://lexidraw.vercel.app` hits Vercel's real http→https redirect and gives `NOT_LEXIDRAW_SERVER` ending 'Point LEXIDRAW_URL at https://lexidraw.vercel.app.' Exit 1.

PR #63, opened 2026-09-24
fix: answer NOT_FOUND when changing or revoking a share that does not exist

### LINKED ISSUE #58: Share change and revoke report success for a user with no share
## What to build

Changing or removing a share for a user who has no share row on that entity answers `200 {"success":true}`, reporting a change that did not happen. A caller cannot tell a real revocation from a no-op, so a script that revokes access cannot confirm it did anything.

Both paths should answer `NOT_FOUND` when the named user holds no share on the entity, the same way every other address that names nothing does. Ownership is already checked, so a foreign entity is correctly a 404 today; this is only about the missing row.

## Acceptance criteria

- [ ] `PATCH /api/v1/entities/{id}/shares/{userId}` for a user with no share row answers `NOT_FOUND`.
- [ ] `DELETE /api/v1/entities/{id}/shares/{userId}` for a user with no share row answers `NOT_FOUND`.
- [ ] Revoking a share that does exist still answers success, and the browser's sharing UI is unaffected.
- [ ] `docs/agent-access.md` says what each answers.

## Blocked by

- None. Found by the QA pass on the agent-access feature; pre-existing, reachable over REST, tRPC, and the browser.

r4iju:
## What

`PATCH` and `DELETE /api/v1/entities/{id}/shares/{userId}` (tRPC `entities.changeAccessLevel` / `entities.unShare`) answered `200 {"success":true}` for a user with no share on the entity. Both now answer `NOT_FOUND` ("Share not found") when no row matched: the update and the delete use `.returning()`, and an empty result throws. A revoke that answers success therefore did revoke something, and repeating it is a 404. `docs/agent-access.md` says what each answers.

The sharing dialog used to roll back silently when a level change failed, and showed "Something went wrong" when a revoke failed. On either error it now rolls back, refetches the share list, and toasts the server's message. A share already revoked in another tab disappears instead of lingering.

Closes #58

## How I verified it

- Tests first: three REST route tests (change on a missing share is 404, revoke on a missing share is 404, and a real share changes, revokes, then a repeat revoke is 404). They failed with 200 before the fix and pass now.
- `bun test` (apps/lexidraw, .env.test): 234 pass, 0 fail. `tsgo --noEmit` is clean. Packages build. The Biome count stays at the 33 baseline.


r4iju:
Review outcome, fixed in a0481f17:

- **The two onError handlers were duplicated, and neither revalidated the dashboard.** Accepted. They now share one `rollBackShare`, which restores the list, invalidates it, calls `revalidateDashboard()` and shows a toast. The toast also appears when there is no optimistic context.
- **The optimistic `temp-id` row had live controls.** Accepted. Change-access and Unshare are disabled on that row until the server answers. The id is now a named constant, `PENDING_SHARE_ID`.

Gate: 234 app tests pass (with `.env.test`), tsgo is clean, and Biome reports nothing for the file.

r4iju:
**UI QA on prod (merged build): passed**

1. Via REST, re-shared the smoke doc with the QA account.
2. Opened the doc's Share dialog in the browser, so the row was on screen.
3. Via REST, revoked the share, which made the row in the open dialog stale.
4. Clicked **Unshare** on the stale row.

Result:
- An error toast says "Share not found".
- The optimistic rollback plus refetch dropped the row. A DOM check found `{"toast":1,"rowsLeft":0}`.
- No unhandled error appeared.

Prod: stale Unshare shows 'Share not found' toast and the row is gone

PR #64, opened 2026-09-24
fix: list only tags that a filter can find

### LINKED ISSUE #59: Tags of soft-deleted entities keep showing in the tag list
## What to build

Deleting an entity is a soft delete: the row keeps a `deletedAt` and its rows in `EntityTags` survive. `GET /api/v1/tags` (and the tRPC procedure behind it) lists tags without excluding trashed entities, so a tag whose only entity is in the trash keeps appearing in the tag list and in the sidebar, and filtering by it returns nothing.

Decide and implement one of: the tag list excludes tags with no live entity, or tag counts report live entities only. Whichever is chosen, a restored entity must come back with its tags, since trash is restorable.

## Acceptance criteria

- [ ] A tag whose only entity is in the trash no longer appears in the tag list.
- [ ] Restoring that entity brings the tag back with it.
- [ ] Tag counts, wherever they are shown, count live entities only.

## Blocked by

- None. Found by the QA pass on the agent-access feature; pre-existing and equally visible in the browser.

r4iju:
## What

`GET /api/v1/tags` (`entities.getUserTags`) read `EntityTags` without looking at the entity. So a tag whose only entity was trashed, or whose entity the owner had since unshared from the caller, stayed in the dashboard tag filter and the article-embed picker, and filtering by it returned nothing. The list now shows a tag only if filtering by it finds something. The query joins `Entities` and requires the entity to be outside the trash **and** owned by or still shared with the caller. Dedupe and sort moved into SQL.

Tag rows are kept, so a restore or a new share brings a tag back. Archived entities still count: `GET /entities?includeArchived=true` lists them, and the tag filter sits next to that toggle. No tag counts exist anywhere. The OpenAPI summary, the CLI's fixture copy of it, and `docs/agent-access.md` now say what the list returns.

Closes #59

## How I verified it

- Tests first. The trash test fails against the old query with `Expected to not contain: "rest-only-in-trash"`. The unshared-entity test (added after review) fails with `Received: [ "rest-unreachable" ]`. Both pass now.
- `bun test` (apps/lexidraw, .env.test): 233 pass, 0 fail. `tsgo --noEmit` is clean. Packages build. The Biome count stays at the 33 baseline.


r4iju:
**Review outcome.** A fresh-context review found nothing blocking. I checked each finding:

- **Accepted and fixed in this PR:** a tag you put on an entity someone shared with you kept listing after the owner unshared it, and filtering by it found nothing. That is #59's failure through a different door, so the list now follows the rule the fix exists for: a tag is listed only if filtering by it finds something. The entity has to be outside the trash **and** owned by or still shared with the caller. The new test was seen failing first (`Received: [ "rest-unreachable" ]`). The OpenAPI summary, CLI fixture and doc are updated. 233 pass, `tsgo` is clean, and the Biome count stays at 33.
- **Declined: an extra test for a tag on both a trashed and a live entity, plus dedupe and order.** It would be green from the start (`SELECT DISTINCT … ORDER BY`, and `Tags.name` is unique), so it would be backfill.
- **Noted, unchanged: children of a trashed directory keep their tags.** This matches how load, search and `list?parentId=` already treat those children, which is as live. Changing what "in the trash" means for descendants is a separate decision.

r4iju:
QA on prod (merged build): I created a doc in the smoke directory and tagged it `qa-64-trash`. `GET /tags` listed the tag (count 1). After `DELETE /entities/{id}` (trash), `GET /tags` no longer lists it (count 0).

PR #65, opened 2026-09-24
fix: hash passwords with salted scrypt and upgrade legacy hashes on sign-in

r4iju:
## Why

Credentials passwords were stored as an unsalted, single-round SHA-256 hex digest and compared with a non-constant-time `===`. If the database leaked, those hashes would be cheap to crack. Hashes could also reach the server logs: the JWT `update` callback logged the full user row, and failed queries were logged with their parameters.

## What changed

- `server/auth/password.ts` adds `hashPassword`, `verifyPassword` and `needsRehash`. New values are `scrypt$N$r$p$salt$hash`, using `node:crypto` scrypt with N=2^17, r=8, p=1, a 16-byte salt and a 32-byte key. Each hash takes about 180 ms and 128 MiB. Comparison uses `timingSafeEqual`. Legacy 64-hex SHA-256 values still verify in constant time. A stored value above today's parameters, or with a key other than 32 bytes, is refused.
- Sign-in (`server/auth/credentials.ts`) rewrites a legacy or outdated hash after a successful login. The rewrite replaces only the value it just verified, so a concurrent change wins. If the rewrite fails, it is logged and the login still succeeds. There is no forced reset and no schema change: `users.password` is already `text`.
- An unknown email, or an account with no password, is checked against a fixed scrypt hash, so response time does not show which accounts exist. Legacy-hash accounts still answer faster until their first sign-in upgrades them.
- Signup stores the new format. No other code writes a password.
- Rehash and signup failures log only a fixed message, the user id and the driver error code. The JWT update callback reads only `config` and no longer logs it.

## Testing

- Tests were written first and seen failing. They cover the hashing module, sign-in against an in-memory DB (legacy upgrade, wrong password, no rewrite of a current hash, unknown email, rehash failures that log no hash), and signup (stores scrypt; a duplicate email logs no hash).
- `bun test`: 244 pass, 0 fail. `tsgo --noEmit` is clean. Packages build. Biome stays at the 33 baseline.
- Checked by hand against a dev server and the dev DB: signup stores scrypt. A seeded SHA-256 user is refused with a wrong password, then signs in with the right one and is upgraded in the DB.


r4iju:
Review outcome: all findings were accepted and fixed in 7b4432b4.

1. **Rehash failure log leaked hashes.** `DrizzleQueryError`'s message quotes the query's parameters. Both the rehash failure and the signup failure now log a fixed `[Auth]` message, the user id (rehash only) and the driver error code or error name, via `server/auth/error-code.ts`. Tests use a SQLite trigger to refuse the update, and a duplicate-email signup, and check that neither log contains a hash.
2. **A failed `hashPassword` failed a correct sign-in.** Hashing and the update now share one try/catch. A test makes `hashPassword` reject and checks that the user is still returned and the stored value is unchanged.
3. **Unbounded stored parameters.** `parse` refuses N > 2^17, r > 8, p > 1, or a key other than 32 bytes. The test uses genuine hashes at each of those settings. Settings below today's still verify, so those accounts can be upgraded.
4. **Unknown-email timing.** A missing account, or one without a password, is verified against a fixed scrypt hash at today's parameters. The test checks that a verify runs, at current cost, for both cases. Legacy-hash accounts answer faster until their first sign-in.
5. **Config logged in the JWT update callback.** The `console.log` of `dbUser` is removed.

Standards: the rehash-on-sign-in rationale now lives only in the `credentials.ts` docblock. The `password.ts` header is trimmed, and its claim that every read and write of `users.password` goes through it is corrected. The SHA-256 test helper is shared (`src/test/legacy-password-hash.ts`). Logs use `[Auth]` throughout.

Gate: packages build 7/7; `tsgo --noEmit` clean; `bun test` 244 pass, 0 fail; Biome 33, matching the baseline.

r4iju:
QA on prod (merged build, Vercel Node runtime), using a new throwaway account (`qa-scrypt-…@example.com`):
- Sign-up through `auth.signUp` stores the password with scrypt at N=2^17 and returns `true`.
- Correct password: 302 to /dashboard with a session (988 ms cold, 697 ms warm).
- Wrong password: 302 to `error=CredentialsSignin` with no session (822 ms).
- Unknown email (3 runs): 302 in 620–644 ms each, the same range as a real account, so the dummy verify runs.
No memory or timeout errors from the 128 MiB scrypt on Vercel.

PR #66, opened 2026-09-24
chore: quiet prerender interrupts, keep Excalidraw off the server, small fixes

r4iju:
## What
- `src/trpc/server.ts` no longer logs Next's own control flow as a procedure error. That covers a prerender ending while `headers()` was pending, a redirect, and `notFound`. Every build printed `[tRPC] Error in procedure: tokens.list` twice with a stack trace. The check uses Next's `unstable_rethrow`, which also follows `cause`. Real failures are still logged.
- The footer said "© 2026 An Lexidraw"; it now says "© 2026 Lexidraw".
- `vercel.json` drops `functions…render-html/route.memory`. Every build warns: "Provided `memory` setting in `vercel.json` is ignored on Active CPU billing."

- The drawing page imported the Excalidraw board directly, so each open evaluated it during SSR, threw `window is not defined`, and fell back to client rendering. It now uses the existing `ssr: false` wrapper, `board-edit-client.tsx`, which nothing had been using.

## Why
Rough edges found during the agent-access QA and prod smoke test.

## How I verified it
- `next-control-flow.test.ts` failed before the module existed. It now passes, covering a prerender interrupt (bare and wrapped in a `TRPCError`), a redirect, and real failures that must still log.
- `bun test` in apps/lexidraw: 249 pass. `tsgo --noEmit`: 0 errors. Biome is clean on the changed files.
- Separately: merged master, with #62, #64 and #65, type-checks clean. The Vercel build does not type-check (`ignoreBuildErrors`), so I checked it locally.
- Excalidraw: on a dev server, opening a drawing used to log `ReferenceError: window is not defined` from `board-edit.tsx`. It now logs nothing, and in a browser the board renders with its toolbar and canvases and no page errors.


r4iju:
Review outcome (1dda3be9):

- **Medium, 'notFound() inside procedures becomes a 500': the premise is wrong; the symptom is real.** `notFound` in `entities.ts` and `drawings.ts` is a local helper that throws `TRPCError({ code: "NOT_FOUND" })`. Next's `notFound` isn't used there. The only `next/navigation` import under `src/server` is `assertAdminOrRedirect`, and that is called from layouts, actions and a route, not from procedures. So no procedure produces Next control flow today, and I dropped the digest-carrying change I had drafted. The symptom does reproduce on prod: a missing document shows 'This page couldn't load' and a missing drawing shows 'Something went wrong'. The cause is that neither page maps a tRPC `NOT_FOUND` to a 404. That fix goes in a separate PR. The docstring no longer claims that Next handles these errors.
- **Low, the test exercises Next's function:** accepted as-is. It pins the digest and `cause` handling across Next upgrades, which is what the filter relies on.
- **Nit, duplicated error path:** fixed. The log now sits under `if (!isNextControlFlow(cause))` with a single `observer.error`.
- **Nit, drawing page catch-all:** covered by the 404 PR.

Gate: 249 tests pass, tsgo 0 errors, Biome clean.

PR #67, opened 2026-09-24
fix: a missing document, drawing, link or folder renders the 404 page

r4iju:
## What
When the entity a page loads is missing, or isn't the caller's to see, the page now renders the app's 404 page instead of a server error. This covers the document, drawing, link (`/urls`) and folder (`/dashboard/[id]`) pages. `entities.load` and `getMetadata` already answer tRPC `NOT_FOUND`. The new `notFoundOr` in `src/trpc/not-found.ts` turns that into Next's `notFound()` and lets any other failure through. The drawing page's catch-all now calls `unstable_rethrow` first, so it no longer swallows the 404. The print page keeps its deliberate redirect to /dashboard.

The document and link pages load the entity before their protected config calls, so a missing entity is a 404 even for a signed-out visitor.

One existing behaviour is unchanged: these pages stream, so the 404 page arrives with HTTP 200 and `noindex`. `/documents/not-a-uuid` does the same on master.

## Why
Found while reviewing #66. On prod, a missing document shows "This page couldn't load. A server error occurred." and a missing drawing shows "Something went wrong".

## How I verified it
- `not-found.test.ts` was red before the change (Expected true, Received false) and is green now. It covers NOT_FOUND → Next's 404 (checked with Next's `isHTTPAccessFallbackError`) and checks that UNAUTHORIZED, INTERNAL_SERVER_ERROR and plain errors pass through unchanged.
- Dev server (port 3047), signed in as a throwaway dev account: `/documents`, `/drawings`, `/urls` and `/dashboard` with a missing id each render "Sorry, that page doesn't seem to exist" with the `NEXT_HTTP_ERROR_FALLBACK;404` digest. A newly created document and drawing still open normally.
- `bun test` 251 pass, `tsgo --noEmit` 0 errors, Biome has no new diagnostics.
- Signed out and signed in, a missing id on `/documents`, `/urls` and `/drawings` renders the 404 page, with no error digest. A new document and a new link still open.


r4iju:
Review outcome (f3abde23):
- **Medium: signed-out visitors got a server error on documents and links.** Accepted. `load` now runs before the protected config calls, so NOT_FOUND wins. Verified on dev for signed-out and signed-in visitors across documents, urls and drawings. The broader bug you spotted is now confirmed on prod: signed-out visitors can't open *public* documents or links, because `getConfig`, `getAudioConfig` and `getTtsCatalog` are protected. That's a separate fix, coming next.
- **Low: the 404 page arrives with HTTP 200.** The existing streaming behaviour is kept and noted in the PR body.
- **Nit: the cause walk continued past the first `TRPCError`.** Fixed, with a test: a NOT_FOUND wrapped in an INTERNAL_SERVER_ERROR now passes through.
- **Nit: unreachable `if (!document)` and `if (!entity)` checks.** Removed.
- **Nit: the test imports a Next internal.** Kept. Checking with Next's own predicate is the point, and a Next upgrade that moves it should fail loudly.

Gate: 252 tests pass, tsgo 0 errors, Biome has no new diagnostics.

PR #68, opened 2026-09-24
feat(cli): address drawings by path and create them in a directory by path

r4iju:
## What
- `drawing get|put|render` take `<id>` or `--path "Dir/Sub/Title"` [`--nth N`], with the same semantics as the doc verbs: reads guess and warn, an ambiguous `put` fails `AMBIGUOUS_PATH`, an id and `--path` together or `--nth` with an id is a usage error.
- `drawing create` takes `--dir <id>` or `--dir-path P`. `--parent` is removed and now fails as an unknown flag (exit 2).
- `EntityKind` includes `drawing`. The `address`/`parent` helpers move from `doc.ts` to `resolve.ts`, where doc, dir and drawing share them. Each drawing verb now accepts only its own flags.
- Usage text, `skills/lexidraw/SKILL.md` and `docs/agent-access.md` are updated.

## Why
Drawings could only be reached by id, and their parent flag was spelt differently from `doc create`'s.

`put` by id has no type guard, unlike `doc delete`: `PUT /drawings/{id}` goes through `findWritableDrawing`, and the `GET` behind `latest` through `findReadableDrawing`; both answer NOT_FOUND for any non-drawing, which a route test now pins.

## How I verified it
- Wrote the tests first in `test/drawing.test.ts` (and switched the `cli.test.ts` create test to `--dir`). All 10 failed: `--dir`, `--dir-path`, `--path` and `--nth` were unknown flags (exit 2), and `--parent` was still accepted (exit 0).
- `bun test` in `apps/cli`: 221 pass, 0 fail.
- `bun x tsgo --noEmit`: clean. `bun run compile`: builds, and `dist/lexidraw drawing create --title x --parent y` exits 2 with `unknown flag --parent`.
- `bun x biome check apps/cli skills docs`: no diagnostics, the same as on master.


r4iju:
Review outcome (c29cb6c5). Nothing was blocking; all non-blocking items are addressed:
1. `parent` and a duplicate `format` are removed from the global flag list.
2. `address()` now takes `Exclude<EntityKind, "directory">`, and its comment is reworded.
3. The PR body now says the GET behind `latest` goes through `findReadableDrawing`.
4. The `drawing create` usage order matches `cli.ts` and SKILL.md. `[--nth N]` stays in the per-command usage only, as for `doc`.
5. An unknown drawing verb now reports `known`, as `doc` and `dir` do.
6. `address()` refuses an unnamed entity before any input is read or request made. This fixes `drawing put --file -` draining stdin first, and the same pattern in `doc`. There's a test for it.
7. New tests: the drawing's own hint through an ambiguous directory, `--nth` not settling a directory for a write, per-verb flags (`get --file` refused), and a route test pinning `PUT /drawings/{doc id}` → 404.
8. `drawing-format.md` shows `--path`, the stale SKILL line is gone, and the stub comment is rewrapped.

Gate: 226 CLI tests pass, 23 v1 route tests pass, tsgo is clean, compile works, and Biome is clean.

r4iju:
**QA on prod (merged `6c025136`, CLI rebuilt with `bun run cli:install`, skill relinked): passed**

| Check | Result |
|---|---|
| `drawing create --title "QA68 drawing" --dir-path smoke-20260924 --file els.json` | created with 5 elements, inside the smoke directory |
| `drawing get --path "smoke-20260924/QA68 drawing"` | same id, elements returned |
| `drawing put --path … --file - --if-unmodified-since latest` | 2 elements, `updatedAt` advanced |
| `drawing render --path … --format svg` | the SVG contains the new label ("QA68" / "v2", wrapped) |
| `drawing get` with no id and no `--path` | `USAGE`: name the drawing … |
| `drawing create --parent abc` | `USAGE`: unknown flag --parent |
| `drawing frob` | `USAGE` with the per-verb usage |
| `drawing create --dir-path nope-dir-68` | `NOT_FOUND`: no directory titled "nope-dir-68" in the root |

I deleted the QA drawing afterwards.

PR #69, opened 2026-09-24
fix: keep an open editor on the stored entity after writes made elsewhere

### LINKED ISSUE #61: An open editor does not pick up writes made over the API
## What to build

A document or drawing open in the browser does not pick up a write made outside the app. The server-side half is done (#60 tags each cached render and every write procedure revalidates it), but a soft navigation through a prefetched link still shows the old content, and that persists even with the page's cache directive removed entirely. What remains is Next's client-side router cache, which no server-side revalidation can reach.

This matters now in a way it did not before: writing from an agent, the CLI, or the MCP endpoint while a tab is open is the ordinary case for this feature, not an edge case.

Close it with a client refresh signal. Two candidates, and the second is probably the right one since the app already has the channel: a router refresh when the tab regains focus or visibility, or the existing realtime channel telling an open editor that its entity changed so it reloads. A reload must not discard unsaved local edits silently.

## Acceptance criteria

- [ ] With a document open in one tab, a write over `/api/v1` or `/api/mcp` is visible in that tab without a manual hard reload.
- [ ] The same holds for a drawing.
- [ ] Unsaved local edits are not lost without the user being told.
- [ ] Navigating back to a listing after an out-of-band write shows the new state.

## Blocked by

- None. Server-side revalidation landed in #60; this is the client half.

r4iju:
## What
An open document or drawing editor now picks up writes made elsewhere, such as over `/api/v1`, `/api/mcp` or the CLI. It never silently throws away unsaved edits.

- **Revision checks.** The editor keeps the revision and content it shows. It checks `entities.revision` (tRPC only, returns just `updatedAt`) on open, on focus and visibility, and every 10 s while visible. When `updatedAt` has moved, it loads the entity and decides in `lib/open-entity-sync.ts`:
  - content unchanged or already shown → adopt quietly
  - no unsaved edits → replace, clear undo, show a notice
  - unsaved edits → ask "Reload" or "Keep mine". Autosave waits while the question stands; a manual save still goes through and answers it. Leaving the editor while it stands asks instead of autosaving. "Keep mine", or closing the prompt, lets autosave save the held edits. "Keep mine" is remembered by content, and later polls don't re-parse or re-ask a question already put.
- **Own saves.** The editor's own saves, read from the `entities.save` mutation cache, don't count as external changes. A save that lands during a check cancels that check. Out-of-order save answers can't move the held revision back. A reload chosen during a save waits for it, for up to 30 s; a save that answers after that is ignored.
- **Adapters.**
  - Lexical (`lexical-edits.ts`): ignores the rewrites Lexical makes by itself after loading, and counts typing in nested editors (captions, sticky notes) as edits. Nested editors are found with mutation listeners on the caption-bearing nodes and let go when their node is removed. The document is serialized only when asked, not per keystroke.
  - Excalidraw (`scene-edits.ts`): compares element `id@version` plus the background colour, worked out once per change. Edits count once the user touches the page anywhere (pointer, key, paste, drop), not only the canvas. A drag that spans a reload stays the user's. A scene the server already stores schedules no save, so a reload is never autosaved back.
- **Server.**
  - `entities.load` also returns `updatedAt`, and both reads share one access check in `server/entities/readable.ts`.
  - Every write of `screenShot*` stamps `thumbnailUpdatedAt`: `entities.update` (custom icons), `snapshot.create`/`update`/`saveUploadedUrl`, the URL distill and the thumbnail job. The thumbnail job no longer moves `updatedAt`, since a picture of the content isn't a new revision of it. Listings cache-bust thumbnails on `thumbnailUpdatedAt`.
- An entity that is deleted or no longer shared is announced once. It is checked again on focus and every 5 minutes, and the notice closes if it comes back.

## Why
Closes #61. The editors read the entity once, at mount:
- `router.refresh()` data never reaches them.
- Navigating back to a page is served from the client cache.
- Nothing tells the browser about writes made through the API.

Rejected alternatives:
- Refreshing page data would destroy unsaved edits.
- Realtime push has nothing on the server that publishes writes to it, and the socket is unauthenticated.

## How I verified it
- Unit tests. Each test of a behaviour this PR changes was seen failing first. The style-change, delete and pan/zoom cases in `scene-edits.test.ts` pin behaviour that already held.
  - `open-entity-sync.test.ts` (16)
  - `open-entity-toasts.test.ts` (3: closing or "Keep mine" resumes autosave, the gone notice)
  - `scene-edits.test.ts` (11: style change, delete, pan/zoom alone, background, a drag across a reload, and the page's input wiring on jsdom)
  - `lexical-edits.test.ts` (5, Lexical on a jsdom root so mutation listeners run, with a real `StickyNode`)
  - `leave-choice.test.ts` (3)
  - `thumbnail-writes.test.ts` (2, router callers on the in-memory database)
  - `thumbnail-src.test.ts` (1)
  - `update-entity-step.test.ts` (1)
- Playwright against a dev server, with the writes made from the CLI or REST. The 13 earlier scenarios, plus:
  - "Keep mine" resumes autosave, for a document and a drawing
  - a custom icon replaced twice through the Thumbnail dialog; the dashboard shows the second one
  - leaving during a conflict (see Known limits)

  All pass except the leave scenario, which master's navigation guard can't exercise.
- Gate:
  - `bun test` 321 pass
  - `tsgo` 0 errors
  - Biome: same findings as master
  - `next build` succeeds

## Known limits (follow-ups)
- Browser saves carry no precondition. So "Keep mine" followed by a save, or a manual save while the question stands, overwrites the other write. The prompt says so.
- The in-app unsaved-changes guard doesn't fire on master: "Go to dashboard" and the back button leave a dirty document without asking, even with autosave off. This PR makes the guard ask during a conflict (unit-tested), but you can't see it in the browser until the guard itself is fixed.
- Custom icons overwrite a fixed blob path. The listing URL now changes, but Vercel Blob's CDN ignores the query and takes a while to pick up an overwrite. If the image optimizer fetches during that window, it caches the old picture under the new URL. This happens on master too. Unique blob paths per upload would fix it.
- Title changes aren't synced.
- Drawing autosave skips changes to neither elements nor background, such as pan or zoom. The viewport is saved with the next real edit.
- `entities.update` still moves `updatedAt` for renames, moves and custom icons. An open editor adopts these with one extra load, because the content is unchanged.


r4iju:
Review outcome. Every item was accepted and fixed in cb70659e. The branch is rebased on master (#66, #67), and the drawing is now loaded client-only through `board-edit-client`.

**Blocking**
- **B1, reload echoed back as an autosave: fixed.** `scene-edits.ts` decides whether an `onChange` holds anything to save. A scene the server already stores needs no save, and that covers the one a replace just showed. `onSyncReplace` cancels any pending debounced save, and a no-op `onChange` no longer calls `markDirty`. Covered by the test "the onChange a replace causes schedules no save", which went red then green. Browser B1: two REST writes about 1 s apart with autosave on. Before, the second write was overwritten (stored magenta, 1 save from the tab). After, blue is stored with 0 saves from the tab.
- **B2, touches outside the canvas: fixed.** A touch is now a capture-phase `pointerdown`, `keydown`, `paste` or `drop` anywhere on the window. Browser B2 inserts from the library sidebar, then an external write lands. Before, the scene was replaced silently. After, it asks and the insert stays.

**Also fixed**
1. **Nested editors: fixed**, though not through the `editors-context` registry. That registry only holds slide boxes, whose edits already reach the root. `lexical-edits.ts` finds every editor nested in a node (`__caption`: image, inline image, sticky, video), watches it, and re-serializes the whole document on a user update. Covered by headless tests with a real `StickyNode`: an edit counts, a merge-tagged pass doesn't, and a sticky added later is watched too.
2. **Autosave answered a standing conflict: fixed.** `holdsSaves()` pauses document and drawing autosave while the question stands, and marks the tab dirty instead. Manual saves still go through and answer the question. The old test that recorded the bad behaviour is replaced. Browser: before, the autosave overwrote the write; after, the write survives and the question stays.
3. **"Keep mine" keyed on the timestamp: fixed.** It is now keyed on content. Tested with the same content under a new `updatedAt`.
4. **`held` moved before `replace` succeeded: fixed.** The replace now runs first, then `held` updates. Tested with a replace that throws, followed by a retry that succeeds.
5. **Reload during a save: fixed.** The reload is deferred. If the save fails, the reload runs; if it lands, there is nothing to reload to. `saveSucceeded` only adopts a newer `updatedAt`. Both are tested.
6. **Toast outlived the editor: fixed.** The effect cleanup dismisses it.
7. **Unused `revision` on the conflict notice: removed.**
8. **Duplicate Excalidraw handle: fixed.** One state value remains, and `DrawingBoardMenu` takes the handle instead of a ref.
9. **Polling after NOT_FOUND: fixed.** `revision` answering NOT_FOUND marks the entity gone, and later checks make no request. Tested.
10. **Thumbnails moving `updatedAt`: fixed.** Before changing it I checked what depends on it:
    - Dashboard ordering is only more accurate without it.
    - Cache tags are revalidated separately.
    - REST/MCP `ifUnmodifiedSince` is actually helped: a finished thumbnail used to break an agent's chained writes with a 409.
    - Thumbnail cache-busting did depend on it. The blob path is fixed and overwritten, and the listing appended `?v=updatedAt`. It now uses `thumbnailUpdatedAt`.

    Tests: `update-entity-step.test.ts` (in-memory database) and `thumbnail-src.test.ts`, both red then green. Not changed:
    - `entities.update`, which bumps `updatedAt` on renames, moves and icons. That's a metadata write with its own semantics; the editor adopts it with one extra load.
    - `generateThumbnailsViaWorker`, which also bumps it but has no callers.

**Gate:**
- `bun test` 276 pass
- `tsgo` 0 errors, from a clean incremental cache (a stale `tsconfig.tsbuildinfo` from before the rebase had hidden one error)
- Biome 33, same as master
- `next build` succeeds

**Browser matrix, 13 scenarios:**
- master: all fail except listing-back
- first revision: B1, B2 and autosave-conflict fail
- now: 13/13 pass

r4iju:
Second review round, in 50b64a04 (rebased onto #75 and #77):

1. **Leaving during a conflict: fixed.** `confirm()` asks instead of saving while a save hold stands (`leaveChoice`, tested). The editors register `holdsSaves` with the provider. In the browser, master's guard doesn't run for in-app navigation at all, so there's nothing to see there yet; this is in Known limits.
2. **Autosave after "Keep mine" or a dismiss: fixed.** `keep()` emits `resumed`, and the editors flush a save when something is unsaved. A reload waiting on a save that never answers goes ahead after 30 s, and that save's late answer is ignored. Tests cover the controller and the toast wiring; Playwright covers it for a document and a drawing.
3. **Thumbnail cache-busting: fixed.** Every `screenShot*` write goes through `thumbnailColumns()`, which stamps `thumbnailUpdatedAt`. Router tests cover `entities.update` and `snapshot.saveUploadedUrl`. Playwright replaces an icon twice through the dialog and the dashboard shows the second. There is a separate issue on master: the Blob CDN can briefly serve an overwritten icon (Known limits).
4. **NOT_FOUND: fixed.** The editor says so once, checks again on focus and every 5 minutes, and dismisses the notice when the entity is back.
5a. **Pointer held across a reload: fixed.** 5b (following the baseline until fonts load): **skipped**. Treating changes as non-edits until an async font event errs the unsafe way: a real edit made in that window would be replaced unasked. Fonts are loaded by the time a check can replace anything.
6. **Background colour: fixed.** It's in the scene key, and saves pass their `appState` through so the baseline matches.
7. **Nested Lexical editors: fixed.** They're now watched through `registerMutationListener` on the caption-bearing nodes and let go on `destroyed`, with a test. Serialization is lazy and cached until the next edit.
8. **Standing question: fixed.** Later polls with the same `updatedAt` return before loading or parsing.
9. **Save precondition: not in this PR.** It stays in Known limits.
10. **Scene key once per change: fixed.** The hook's own `onChange` subscription is gone, and both boards forward Excalidraw's `onChange`.

With #77 on master (`compilationMode: "infer"`), this PR's `"use no memo"` directives were no-ops and are removed.

r4iju:
Prod QA on the merged build (0c92b6c2):
1. Created a QA document via the CLI and opened it in the editor. It showed "First paragraph from the CLI."
2. With the editor still open, ran `lexidraw doc append <id> --text "Second paragraph appended by an agent."`, which returned updatedAt 03:36:26Z.
3. When the tab regained focus, the open editor showed both paragraphs within about 2 seconds, with no reload and no page errors. Before this change, the editor kept its stale copy.

The QA document was deleted afterwards.

PR #71, opened 2026-09-24
feat(cli): delete a drawing by id or path

r4iju:
**What:** adds `lexidraw drawing delete <id|--path P>`. Until now the CLI could create, write and render a drawing, but the only way to remove one was the raw `api DELETE /entities/{id}`. I hit this while cleaning up after the #68 prod QA.

**How:**
- `doc delete`'s resolve, kind-check and DELETE logic moves into `src/delete.ts` as `deleteEntity(context, args, kind)`, which both nouns use.
- `DELETE /entities/{id}` takes any entity, including a directory with everything under it. So an id is checked against the noun before anything is deleted ("\"x\" is a directory, not a drawing"), and a `--path` only matches the noun's own kind.
- An ambiguous `--path` is refused, as for every other write.
- The command map (`schema "drawing delete"`), `--help`, SKILL.md and agent-access.md are updated.

**How I verified it:**
- I wrote the tests first in `test/drawing.test.ts`: the tests below failed (4 fail), then passed.
  - delete by path, which spares the document of the same title;
  - delete by id;
  - an id naming a directory or a document is refused and nothing is deleted;
  - an ambiguous path is refused.
- `bun test` in apps/cli: 230 pass. `tsgo --noEmit -p apps/cli`: clean. Biome: clean.


r4iju:
Review outcome (fresh-context reviewer, nothing blocking). Each finding checked against the code:

1. **`agent-access.md:230` still listed `drawing get|put|create|render`.** Fixed: the list now includes `delete`.
2. **A drawing that is shared with you, not yours, answered a bare "Entity not found", and the skill sent the reader to look for a path problem.** Fixed in the CLI rather than the server. The entity was just read or listed, so a NOT_FOUND from the DELETE means it isn't the caller's. The CLI now says `"<id>" was not deleted: only its owner can move it to the trash` and keeps the server's details. The SKILL's NOT_FOUND row and the Drawings section now say delete goes to the trash and is owner-only. There's a test for both address forms, using a `shared` row in the stub that the stub refuses to delete.
3. **Help left out `--nth` for delete.** Fixed in the drawing USAGE. There's now a test that `--nth 2` deletes the older of two drawings with the same title.
4. **`doc delete` was only tested against a directory id.** It now loops over a directory and a drawing.
5. **The comment said a directory delete "includes everything under it".** Reworded: deleting a directory takes everything under it out of sight (the server only stamps the directory itself).
6. **The kind check downloads the whole drawing.** Not changing it. No lighter endpoint exists, and it only runs for delete-by-id.

`bun test` (apps/cli): 232 pass. tsgo and Biome: clean.

r4iju:
**QA on prod (merged `6f9a3ede`, CLI rebuilt): passed**

| Check | Result |
|---|---|
| `drawing create --title "QA71 drawing" --dir-path smoke-20260924` | created |
| `drawing delete <directory id>` | refused: `"8bf3…" is a directory, not a drawing`, nothing deleted |
| `drawing delete <document id>` | refused: `"55ab…" is a document, not a drawing`, nothing deleted |
| `drawing delete --path "smoke-20260924/QA71 drawing"` | `{ "id": "d104…" }` |
| `drawing get` on the same path | NOT_FOUND. The listing shows only the smoke doc and the smoke drawing |

Not checked on prod: the owner-only message. It needs a drawing someone else has shared with the token's owner. The stub test covers it for both the id and the path form.

PR #72, opened 2026-09-24
fix: name the dashboard's icon-only buttons

r4iju:
**What:** gives every control on the dashboard and the header an accessible name. On prod a DOM audit of the dashboard found 7 controls a screen reader announces as just "button" or "link":
- the account menu;
- the grid and list layout links;
- each row's tag button;
- each row's ⋯ menu.

Chromium's accessibility tree found three more:
- the tag filter and sort comboboxes, which had a value but no name;
- the draggable cards, each named by its entire content.

**How:**
- An sr-only span, as the theme toggle already does:
  - "More actions for X", "Tags for X", "Account menu", "Grid view"/"List view", "Remove tag X".
  - The mobile-only Favorites and Archived icons are named by text that is `sr-only` below `md`.
- `aria-label` where the element can't be named from its content:
  - the comboboxes, "Filter by tags" and "Sort by";
  - the draggable cards, "Move X";
  - the rename controls, "Title", "Save title" and "Cancel renaming".
- `aria-current="true"` on the active layout link and on the Favorites and Archived filters while they are on.
- `AppIcon` set its own `aria-label` and `className` after `{...props}`, which replaced the header's "Go to startpage" with "App icon" and silently dropped its classes.
  - It now merges `className` and takes its `<title>` from the label.
  - The two header callers drop the classes that were never applied, so the logo looks the same (24px, as before).

**How I verified it:**
- `build:packages`, `tsgo --noEmit -p apps/lexidraw`: clean. `bun test` in apps/lexidraw: 256 pass. Biome on the touched folders: only the diagnostics master already has.
- In the browser, signed in on the dev server, I read Chromium's accessibility tree over CDP (`Accessibility.getFullAXTree`).
  - Unnamed interactive nodes at 1280px and 390px, in both layouts: none apart from the dev-only overlay's ✕, which isn't this app's code.
  - Names are as listed above. The tag filter reads name "Filter by tags" with value "0" on mobile. The rename input is "Title".
- The layout is unchanged. The purple boxes are dev-only react-scan overlays.

Mobile dashboard after (unchanged layout)
Desktop dashboard after (unchanged layout)

**Follow-ups (editor, out of scope):** icon-only buttons in the document editor's plugins are still unnamed. They are in the comment, floating-link, table-action, TTS, LLM chat, autocomplete and slide plugins. The shared `EllipsisIcon` hard-codes "Ellipsis icon" as its label.


r4iju:
Review outcome (fresh-context reviewer). Each finding checked in Chromium's accessibility tree:

1. **Blocking: the tag filter combobox was unnamed.** A `role="combobox"` takes no name from its content, so the sr-only span only fed its value. The reviewer was right, and my first check read content, not the computed name. Fixed with `aria-label="Filter by tags"`; the span is gone. CDP now reports name "Filter by tags" with value "0" on mobile. The same audit also found the sort combobox unnamed ("Sort by") and the draggable cards named by their entire content ("Move X"). Both are fixed. Biome can't see the role dnd-kit spreads, hence the reasoned `biome-ignore`.
2. **`aria-current="page"` → `"true"` on the layout links.** Agreed, they're two views of one page. Changed.
3. **The Favorites and Archived toggles don't expose their state.** Fixed with `aria-current="true"` while on.
4. **`AppIcon` dropped the caller's `className`, and its `<title>` said "App icon".** It now merges `className` and titles itself from the label. Applying the header's never-used `size-full …` classes would have made the logo fill the button, so the header callers drop them. The logo is still 24px.
5. **The rename input had no label.** It now has `aria-label="Title"`.

256 tests pass. tsgo is clean. Biome is the same as master on the touched folders.

r4iju:
**QA on prod (merged `4a59d943`, deployed): passed**

I read Chromium's accessibility tree over CDP on the prod dashboard, signed in, in the smoke directory.

| Viewport, layout | Interactive nodes | Unnamed |
|---|---|---|
| 1280px, list | 27 | **0** (before: 7) |
| 390px, grid | 25 | **0** |

Names on prod:
- "Go to startpage", "Account menu" and "Toggle theme"
- "Filter by tags" and "Sort by" (comboboxes, with their values)
- "Favorites", "Archived", "Grid view" and "List view"; the active view has `aria-current=true`
- "Move Smoke doc", "Tags for Smoke doc" and "More actions for Smoke doc"

The mobile layout is unchanged:

Prod dashboard at 390px after #72, layout unchanged

PR #73, opened 2026-09-24
fix: open public documents and links for signed-out visitors

r4iju:
## What
- Settings reads in `config` (`getConfig`, `getAudioConfig`, `getAutoSaveConfig`, `getTtsConfig`, `getArticleConfig`, `getAutocompleteConfig`) answer a new account's defaults when there is no session. Every settings write stays protected. `getTtsCatalog` stays protected.
- The document and link pages read the session on the server. The document editor gets it through a `SignedInProvider`, so account-only tools (the AI assistant, autocomplete, the TTS toolbar and AI image status) are right on first paint and are not offered to a visitor. The link page skips the TTS catalog and hides audio generation for a visitor. Saved audio still plays.
- `/api/llm/agent` answers 401 without a session and 404 for a document the caller cannot read. It used to throw, which Next turned into a 500.
- Removes the unused `url-editor.tsx` and `url-editor-client.tsx`.

## Why
A signed-out visitor to a public document or link got a server error. `entities.load` serves public entities to anonymous callers, but the pages also called protected `config` reads, which threw `UNAUTHORIZED`. Serving defaults on the read side keeps any viewer page working without each caller having to check for a session. The catalog is not a user setting: building it calls Google and the TTS sidecar with the app's credentials, and the sidecar returns paths from its own disk.

## How I verified it
- `config.test.ts` uses its own database, holding one user with a non-default value in every section, and asserts literal defaults for a visitor. The six visitor-read tests failed with `UNAUTHORIZED` before the fix. With the lookup changed to not filter by the caller, the visitor got the stored values and all six failed again. Guard tests check that visitors are still refused every write and the catalog.
- `api/llm/agent/route.test.ts` checks for 401 without a session and 404 for another user's private document. Both tests failed before the fix, because the route threw.
- Browser, signed out: the public READ doc, public EDIT doc and public link render with no page errors and no failed calls. The EDIT doc makes no autocomplete calls and shows no LLM chat trigger, even with autocomplete on and a chat left open in localStorage. Private entities still give a 404.
- Browser, owner: on a hard load, the TTS toolbar enters the DOM in the same mutation as the toolbar, so nothing pops in. The owner's view is otherwise unchanged.
- `bun test` (267 pass), `tsgo --noEmit` and `biome check` all pass. Biome shows 30 warnings and 3 infos, the same as master.


r4iju:
Review outcome, item by item:

1. **Sign-in detected inconsistently / TTS pop-in: fixed.** The document page computes `signedIn` with `auth()` on the server. It works under `"use cache: private"` and was checked for both owner and visitor. `DocumentEditor` provides it through a small `SignedInProvider`. A context was needed rather than prop threading, because `SlideDeckEditor` also renders `ToolbarPlugin`. `ImageGenerationProvider` takes an explicit prop, and the dashboard passes `true`. `useSession` now stays only in the audio player, to gate saving the rate. On an owner's hard load, the TTS controls appear in the same mutation as the toolbar (0 ms apart). The `document-editor.tsx` change merges cleanly with #69 (checked with `git merge-tree`).
2. **Tests wouldn't catch the leak: fixed.** The config tests now run on their own database (new `createTestDatabase()` helper), so the settled user is the row an unfiltered lookup would find. They assert literal defaults. With `where: userId ? eq(...) : undefined`, all six visitor reads went red (for example, `preferredPlaybackRate` came back 1.5 instead of 1). They went green again once the change was reverted.
3. **Agent route 500: fixed.** It now answers 401 without a session and 404 when the caller cannot read the document, following the repo's not-found-rather-than-forbidden convention from #67. There is a new route test (red before, green after), and a live anonymous POST gets a 401.
4. **Stale comments: fixed.** Both pages now say "A missing document/link, or one this caller may not read, is a 404."
5. **Autocomplete and LLM chat for visitors: fixed.** Both are gated on `signedIn`. The chat sidebar also stays shut when an earlier sign-in left it open in localStorage. A signed-out EDIT doc with autocomplete turned on makes 0 autocomplete calls and shows no chat trigger.
6. **ArticlePreview queries and dead URL editor: fixed.** `getTtsConfig` and `getArticleConfig` now use `enabled: canGenerateAudio`. Nothing imports `url-editor.tsx` or `url-editor-client.tsx` (grep across the repo, including dynamic imports and string paths), so both are deleted.

r4iju:
**QA on prod (merged `474d2c73`, deployed): passed**

Setup: in the smoke directory I created a throwaway document and link, with the link distilled from example.com, and set both to public READ. Then I opened them in a fresh browser context with no session.

| Check | Result |
|---|---|
| Public document, signed out | content shown; no page errors; no failed tRPC calls; no LLM chat trigger |
| Public link, signed out | article shown; no failed tRPC calls; no Listen button |
| Same link, owner | article shown; Listen button present |
| `POST /api/llm/agent` with no session | **401** (was a 500) |

Found along the way, and not from this PR: React #418, a hydration text mismatch, on every distilled link page, for owners too. It comes from `toLocaleString()` in ArticlePreview and dates from when the reader landed. The fix is #75.

Prod, signed out: public document renders
Prod, signed out: public link renders, no Listen button

I'm deleting the QA entities after #75's QA.

PR #75, opened 2026-09-24
fix: render timestamps without a hydration mismatch

r4iju:
**What:** timestamps rendered by client components during SSR no longer cause React error #418. This covers the distilled link page, the API token list, the admin tokens, users, entities, LLM usage and thumbnail-job tables, and the article block in documents. It also fixes two ways a link page could crash: a malformed `elements` value or an invalid date.

**Why:** these sites formatted dates with `toLocaleString()` or date-fns `format`. The server uses its own zone and locale and the browser uses the reader's, so the text differed at hydration. On prod every distilled link page threw #418, for owners and visitors alike (found during the #73 QA). Separately, `new Date("yesterday").toISOString()` throws, and the link page cast `JSON.parse(entity.elements)` without checking it.

**How:**
- `LocalTime` (`components/ui/local-time.tsx`) renders a fixed UTC form on the server and swaps in the reader's form after hydration, through `useSyncExternalStore` with a `null` server snapshot. `format="date"` / `"datetime"` keeps the admin tables' `yyyy-MM-dd[ HH:mm:ss]` look. The server shows a UTC form of the same shape. An invalid date renders nothing.
- The link page parses `elements` with a zod schema, and each field that doesn't fit falls back to absent.
- The module opts out of the React Compiler (`"use no memo"`). In `compilationMode: "all"`, dev builds give every function a `_c()` cache for the hot-reload reset. Production builds don't. `subscribe` runs outside render, so that cache throws "Invalid hook call". The formatters run inside `getSnapshot`, where they clash with `LocalTime`'s own cache. I confirmed both on a fresh `.next`.

**How I verified it:**
- I wrote the tests first and watched them fail. A jsdom test renders to a string in UTC, hydrates in Asia/Tokyo, and collects `onRecoverableError`. Against the old `toLocaleString()` markup it fails with "Hydration failed because the server rendered text didn't match the client". Other tests cover invalid dates, the date and datetime forms, and the zod fallbacks.
- `bun test`: 278 pass. tsgo: clean. Biome: nothing new; the one `useOptionalChain` hint in ArticlePreview is also on master.
- Dev server with a fresh `.next`, in a browser context set to Asia/Tokyo and sv-SE:
  - On the link page, the SSR HTML has `<time dateTime="2026-09-24T02:45:35.932Z">2026-09-24 02:45 UTC</time>`, and the page then shows `2026-09-24 11:45:35`.
  - On `/settings/tokens`, the SSR HTML has `2026-09-25 02:47 UTC`, and the page shows `2026-09-25 11:47:14`.
  - Neither page raised a hydration error or a compiler-cache warning.
  - I couldn't load the admin tables because the dev user isn't an admin. They're client tables rendered by async server pages, so they run during SSR.


r4iju:
How each review finding was handled (ba812b35):

1. **An invalid date crashes the page. Fixed.** `LocalTime` still calls the hook, then renders nothing when the value doesn't parse. `new Date("yesterday").toISOString()` no longer runs.
2. **`distilled` comes from an unchecked cast. Fixed.** `parseLinkElements` (`urls/[urlId]/link-elements.ts`) validates `elements` with zod, with `.catch` on every field. A wrong-typed field, a non-ISO `updatedAt`, a non-object or bad JSON all degrade to absent.
3. **The tests only check server markup. Fixed.** A jsdom test renders to a string in UTC, runs `hydrateRoot` in Asia/Tokyo, and collects `onRecoverableError`. It asserts there are no errors and that the text swaps to the local form. With the old `toLocaleString()` markup swapped back in, it fails with React's hydration-mismatch error.
4. **The same #418 bug exists elsewhere. Fixed.** Moved `settings/tokens`, admin tokens, admin LLM usage (default form), admin thumbnail-jobs (`format="datetime"`), and admin users and entities (`format="date"`). Each is a client component rendered during SSR by an async server page. The ArticleBlock metadata line moved too, for consistency. I left the dashboard's relative times alone.
5. **The `"use no memo"` comment is inaccurate. Partly right.** A standalone production compile does hoist `() => {}` to `_temp` with no cache. But in dev (`NODE_ENV=development`), the compiler turns on `enableResetCacheOnSourceFileChanges` and gives every compiled function a `_c()` slot.
   - Removing the directive on a fresh `.next` brought back "Invalid hook call" on the link page.
   - That also exposed a second case: the module-level formatters, which `getSnapshot` calls during render, raised "Expected a constant size argument for each invocation of useMemoCache".
   - The directive now covers the whole module, and the comment states that reason. The compiler mode itself changes in a separate PR.

r4iju:
Prod QA on the merged build (39d7603a), link page `/urls/598b8f65…`:
- Server HTML: `<time dateTime="2026-09-24T01:52:02.439Z">2026-09-24 01:52 UTC</time>`
- After hydration (Asia/Tokyo): `<time datetime="2026-09-24T01:52:02.439Z">9/24/2026, 10:52:02 AM</time>`
- No page errors or console errors. Before the deploy, the same page logged React #418.

PR #77, opened 2026-09-24
fix: compile only components and hooks with the React Compiler

r4iju:
## What
Switches the React Compiler from `compilationMode: "all"` to `"infer"`, so only components and hooks get a memo cache.

## Why
In `"all"` mode every top-level function got `const $ = _c(n)`. `_c` is `useMemoCache`, a hook, so any helper that ran outside render threw "Invalid hook call" (React #321). In production, right-clicking in the document editor threw and no menu appeared. The same bug also hit `pasteFromClipboard`, `getDOMSelection` (inline-image drag-drop) and the colour picker's `getTouchPoint`/`isInvalid`, and it caused the `LocalTime` subscribe crash in #75. Commit 7450664f patched one instance by hand; this removes the whole class.

## How
- The compiler options now live in `apps/lexidraw/react-compiler.ts`, shared by `next.config.ts` and the new test.
- `src/test/react-compiler.test.ts` runs every app source file through Next's bundled Babel with the compiler and those options. It applies Next's own SWC pre-check (a file with no JSX and no `useX()` call never reaches the compiler) and fails if any function other than a component or hook gets a cache, compiled as `next dev` does. Under `"all"` it lists 33 functions, including all five above. Under `"infer"` it lists none.
- `FolderVisual` becomes `folderVisual`. A hook calls it as a plain function, and only for folders, so as a component it would get a cache that the hook then calls conditionally.
- `LocalTime` drops the module-wide `"use no memo"` it gained in #75. It does nothing under `"infer"`.

## How I verified it
- Ran `bun run test:ts`, `bun run test` (268 app tests) and `bun run lint` (no new diagnostics): all pass.
- `next build --turbopack` (NODE_ENV=production) exits 0.
- Dev (Turbopack): right-clicking in a document opens the menu (Copy / Cut / Paste / Paste as Plain Text / Delete Node) with no page errors.


r4iju:
Review adjudication (fresh-context reviewer):
1. **`"use memo"` has no effect: confirmed, fixed.** `isReactCompilerRequired` returns false for `shortcuts.ts` and `tools/web.ts`, so Next never compiles them, in `"all"` mode either. Directives removed. The shortcuts effect re-registering on every render already happens on master and is harmless, so it's left alone.
2. **The test doesn't apply Next's file filter: fixed.** The test now gates each file on SWC's `isReactCompilerRequired`. Under `"all"` it still names all five known crash sites (33 offenders in total).
3. **`nameOf` misreads a memo/forwardRef-wrapped default export: fixed.** A wrapped anonymous function now counts as a component.
4. **`FolderVisual` is called conditionally: fixed** by renaming it to `folderVisual`, so it gets no cache.
- **Nit, the `react-compiler.ts` comment: fixed.** It now says why the file exists.
Gate: tsgo passes, 279 app tests pass, Biome is clean.

r4iju:
Prod QA on the merged build (69e33b92): right-clicking in the editor on a QA document opens the context menu (Copy / Cut / Paste / Paste as Plain Text / Delete Node), and the page throws nothing. Before this change the same right-click on prod threw React #321 and no menu appeared.

PR #81, opened 2026-09-24
fix: rate-limit sign-in, guard saves and leaving, name editor buttons, fresh custom icons

### LINKED ISSUE #70: Credentials sign-in has no rate limit
`authorizeCredentials` (`apps/lexidraw/src/server/auth.ts`) accepts any number of attempts per email or IP. Two consequences:

- **Brute force**: nothing slows password guessing against a known email.
- **Cost amplification**: since #65, every attempt runs scrypt with N=2^17 (about 128 MB, about 100 ms of CPU), including the dummy verify for unknown emails. A burst of sign-in requests therefore translates directly into function memory and CPU.

**Proposal:** a fixed-window limit keyed by IP and by normalised email, checked before scrypt runs, returning the same generic failure as a wrong password. Storage options:

- a small Turso table, which needs no new infra but costs one write per attempt;
- Vercel KV/Upstash, which is cheaper per hit but adds a dependency.

**Acceptance:** the N+1th attempt within the window never reaches scrypt, is logged by error code only, and a test covers it.

### LINKED ISSUE #74: Document editor has icon-only buttons with no accessible name
#72 named every control on the dashboard and the header. The document editor's plugins still have icon-only buttons that screen readers announce as just "button". A rough grep found candidates, still to be confirmed in the accessibility tree:

- `CommentPlugin/index.tsx` (~437, 520, 649, 684). The ⋯ buttons read "Ellipsis icon", because `components/icons/ellipsis.tsx` hard-codes that `aria-label`.
- `FloatingLinkEditorPlugin/index.tsx` (~287, 300, 325, 343)
- `TableActionMenuPlugin/index.tsx` (~724)
- `TtsToolbar.tsx` (~460), `PlayFromHereButton.tsx` (~335)
- `LlmChatPlugin/ui/{slide-generation-form,reset,message-input}.tsx`
- `AutocompletePlugin/LLMWidget.tsx` (~38)
- `SlideNode/SlideDeckEditor.tsx` (~1635), `SlideNode/SlideModal.tsx` (~106)
- `app/urls/[urlId]/ArticlePreview.tsx` (~585)

Also, the dashboard's per-row "Edit title" button should become `Edit title of X`.

**Check:** with a document open, `Accessibility.getFullAXTree` over CDP shows no unnamed button, link or combobox. Name comboboxes with `aria-label`, since they take no name from their content.

### LINKED ISSUE #76: Browser saves carry no precondition, so autosave can overwrite an API write
`entities.save`, which the browser uses for documents and drawings, accepts any write. The REST/CLI writes have `ifUnmodifiedSince`, but browser saves don't. Two consequences:

- **While a user is editing,** autosave (1 s debounce) usually lands before the open-editor poll (10 s, #69) sees an external write. The tab's save then silently replaces what an agent just wrote.
- **If the server applies two of the tab's own saves in reverse order,** the tab believes the newer one is stored.

#69 narrows the first case: it asks before replacing unsaved edits, and pauses autosave while that question stands. It can't close it, because the tab only learns about the external write on the next poll.

**Proposal:** `entities.save` takes the `updatedAt` the tab holds, as `ifUnmodifiedSince`, and answers CONFLICT when it is stale. The editor treats CONFLICT exactly like a poll that found an external write, and asks with Reload / Keep mine. On Keep mine it re-sends with the new revision. That also removes the reliance on wall-clock comparisons across server instances.

**Acceptance:**
- An agent write that lands between two keystrokes is never overwritten without the question.
- A router test covers a CONFLICT on a stale save.

### LINKED ISSUE #79: Leaving a document with unsaved edits doesn't prompt
"Go to dashboard" and the browser back button both leave a document that has unsaved edits without asking, even with autosave off. The edits are lost.

Found while QA'ing #69 and reproduced on master, so it isn't caused by that PR. #69's new rule that a pending conflict asks before leaving is covered by unit tests (`hooks/leave-choice.ts`), but it can't be seen in the app until the navigation guard runs on client-side navigations.

**Expected:** with unsaved edits, an in-app navigation or back navigation asks first, and so does a tab close.

### LINKED ISSUE #80: A replaced custom icon can stay stale for 30 days
A custom icon upload overwrites a fixed Vercel Blob path. Blob's CDN ignores the query string and takes a while to pick up the overwrite. If the Next image optimizer fetches the icon during that window, it caches the old picture under the new `?v=` URL for 30 days.

Seen once in dev while QA'ing #69.

**Fix:** give each upload a unique blob path, such as an id-and-timestamp suffix, and store that URL. Delete the previous blob afterwards.

r4iju:
Fixes the five issues filed during the agent-access work.

## What changes

- **#70 Credential sign-in is rate limited.**
  - Attempts are counted per email and per IP (IPv6 by /64) in a new `SignInAttempts` table, using one upsert per attempt.
  - Past either limit, sign-in is refused like a wrong password before any scrypt runs.
  - The bucket-cleanup cron purges old windows.
  - If the count can't be taken, the sign-in goes ahead: a counter outage mustn't become a sign-in outage.
- **#74 The editor's icon-only buttons have accessible names,** and so does the document's text area. A headless accessibility-tree check of the dashboard and a document finds no unnamed buttons, links, comboboxes or text areas.
- **#80 A replaced custom icon no longer stays stale.**
  - Every thumbnail upload gets its own blob path.
  - The replaced blob is deleted, unless another row still shows it.
  - `snapshot.saveUploadedUrl` checks write access and accepts only the entity's own `thumbnails/<id>/` uploads on our blob host.
  - Shared editors may upload.
  - The `?v=` cache-busting and the unused `snapshot.create/update` are removed.
- **#76 Browser saves carry a precondition.**
  - `entities.save` takes an optional `ifUnmodifiedSince` and answers `CONFLICT` (409) when the revision has moved on.
  - Each tab sends one save at a time; later saves collapse into the newest content.
  - A save times out after 30 s.
  - A refused save gets the same Reload / Keep mine question as an external write found by polling.
  - A save that was stored but whose answer was lost is recognised as the tab's own write.
  - While a collaborator is connected and the editor has edits, a changed revision is absorbed rather than asked about.
- **#79 Leaving with unsaved edits asks first.**
  - Why the old guard failed: the menus pushed through the raw router, and the editor's popstate listener ran after Next had already unmounted the page.
  - One app-wide guard now handles link clicks, app navigations (`leaveThen`), back/forward and tab close. Its listeners go in ahead of Next's.
  - Each editor registers only what it would lose.

## Deploy

`SignInAttempts` (migration `0019`) is created in prod right before merge. Sign-in works without it, but the limit only takes effect once it exists.

## Verification

- `bun test` passes: 355 tests, 0 failures. Every new test was seen failing first.
- Biome shows 0 errors, and `tsgo` is clean for `apps/lexidraw` and `apps/cli`. `next build` exits 0.
- A headless check against a local production build:
  - Back with unsaved edits asks and keeps the URL, and "Stay" keeps the text.
  - The menu's "Go to dashboard" closes the menu first, then asks, and "Leave" navigates.
  - The browser logs no page errors.
- Two fresh-context reviews. All their findings are fixed with tests:
  - a save loop after unmount;
  - a stuck leave question;
  - hung saves;
  - lost save answers;
  - collaboration ping-pong;
  - a peers flag that never cleared;
  - a clean peer that was never reloaded;
  - the guard re-registering while its question was up;
  - the thumbnail URL checks;
  - the limiter failing closed.

## Known limits

- Live collaboration broadcasts whole states, last write wins. So while a peer is connected and the editor has edits, an API/CLI write can be overwritten by the next save. This is documented in `open-entity-sync.ts`.
- Browsers without the Navigation API treat every back/forward as one step back, and ask on a hash-only back.

Closes #70
Closes #74
Closes #76
Closes #79
Closes #80


r4iju:
## Prod QA (5aaca76a)

Checked from the terminal only, with no browser, against https://lexidraw.vercel.app.

- **Prod database:** `SignInAttempts` and its index were created before the merge with `IF NOT EXISTS`. The table was empty at the time.
- **#76:** tested on a throwaway drawing through `/api/v1` with a CLI token.
  - A `PUT /entities/{id}` with a stale `ifUnmodifiedSince` gets **409** `CONFLICT` ("Drawing was modified at …; re-read it and retry with the new updatedAt").
  - The same save with the current `updatedAt` gets **200**.
  - The drawing was deleted afterwards.
- **#70:** one wrong-password credentials sign-in created one email counter and one IP counter in `SignInAttempts` (hashed keys, count 1 each). The cleanup cron purges them.
- **#79:** checked headless against a local production build of this branch.
  - Back with unsaved edits asks and keeps the URL, and "Stay" keeps the text.
  - The menu's "Go to dashboard" closes the menu, then asks, and "Leave" navigates.
  - There are no page errors.
- **#74:** the headless accessibility-tree check on the same build finds no unnamed controls on the dashboard or in a document.
- **#80:** covered by server tests only. Checking it on prod would mean uploading and replacing a real icon.

